PR #2, opened 2025-03-19
Add support for undecorate-round window style

liaowang11:
Added frame parameter for undecorated round window style, based on this patch: 
https://github.com/d12frosted/homebrew-emacs-plus/blob/master/patches/emacs-30/round-undecorated-frame.patch

### DIFF
--- a/src/frame.c
+++ b/src/frame.c
@@ -976,6 +976,7 @@ make_frame (bool mini_p)
   f->horizontal_scroll_bars = false;
   f->want_fullscreen = FULLSCREEN_NONE;
   f->undecorated = false;
+  f->undecorated_round = false;
   f->no_special_glyphs = false;
 #ifndef HAVE_NTGUI
   f->override_redirect = false;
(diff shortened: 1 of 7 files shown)


jdtsmith:
Thanks for this contribution.  I'm concerned that it affects the `nsterm.h` and `nsfns.m` files, which are not used by this build.  This would make maintenance harder as upstream evolves.  Is this patch automatically applied by the home-brew recipe?  Can you post a screenshot of how it appears?

shuxiao9058:
image

liaowang11:
> Thanks for this contribution. I'm concerned that it affects the `nsterm.h` and `nsfns.m` files, which are not used by this build. This would make maintenance harder as upstream evolves. Is this patch automatically applied by the home-brew recipe? Can you post a screenshot of how it appears?

Yes, I should not have changed the ns files when porting it, but thats an easy fix. As a matter of fact, this [patch](https://github.com/d12frosted/homebrew-emacs-plus/blob/8364d359bd0f7cf989da68005fc1a3f3989498d2/Formula/emacs-plus%4029.rb#L96) is just applied by homebrew before building which is not included in upstream, and that's the reason I'm not sure if this patch should be included in the repository. You can see the screenshots from emacs-plus here: https://github.com/d12frosted/homebrew-emacs-plus?tab=readme-ov-file#no-titlebar

jdtsmith:
OK in that case I think I'd prefer to keep this as a patch on top, since I'm not sure what advantage if offers (my windows already look like that), and I want to keep compatibility with upstream as much as possible. 

See [this home-brew tap clone](https://github.com/pkryger/homebrew-emacsmacport-exp) that sources from this repo; possible to apply the patch in that way.

Issue #6, opened 2025-03-25
native-lisp link was wrong


alanshutko:
I built it following the instructions in README.md. 

When I tried to run it after linking native-lisp, it failed. I had to change the symlink:

```
[15:51:12] hela:mac ats$ ln -s ../native-lisp Emacs.app/Contents/
[15:51:13] hela:mac ats$ ./Emacs.app/Contents/MacOS/Emacs
Error using execdir /Users/ats/Builds/emacs-mac/mac/Emacs.app/Contents/MacOS/:
emacs: dlopen(/Users/ats/Builds/emacs-mac/mac/Emacs.app/Contents/MacOS/../native-lisp/30_1_50-dc1cf33b/preloaded/window-0d1b8b93-70443922.eln, 0x0001): tried: '/Users/ats/Builds/emacs-mac/mac/Emacs.app/Contents/MacOS/../native-lisp/30_1_50-dc1cf33b/preloaded/window-0d1b8b93-70443922.eln' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/Users/ats/Builds/emacs-mac/mac/Emacs.app/Contents/MacOS/../native-lisp/30_1_50-dc1cf33b/preloaded/window-0d1b8b93-70443922.eln' (no such file), '/Users/ats/Builds/emacs-mac/mac/Emacs.app/Contents/MacOS/../native-lisp/30_1_50-dc1cf33b/preloaded/window-0d1b8b93-70443922.eln' (no such file)
[15:51:16] hela:mac ats$ ln -sf `pwd`/../native-lisp Emacs.app/Contents
```

jdtsmith:
The hope is you don't actually need to do this step if you `make install`.  Did you check?

alanshutko:
With make install it works.

PR #7, opened 2025-03-25
Tweak readme to be more clear

alanshutko:
Correct typo in make install. Clarify that linking native-lisp is not required if doing make install.

Inspired by #6 .


jdtsmith:
Thanks.

PR #21, opened 2025-06-14
Add full-featured Window menu with tiling window manager support

rymndhng:
What works:
- native keyboard shortcuts

What's meh:
- coding practises? first time writing obj-c.
- placement of window vs help (non-native)
- emacs keybinds has priority over mac bindings

### DIFF
--- a/lisp/term/mac-win.el
+++ b/lisp/term/mac-win.el
@@ -91,6 +91,12 @@
 (defvar mac-ts-active-input-overlay)
 (defvar mac-frame-tabbing)
 
+
+;; Define the macos-specific Window menu and add it to the global menu-bar map
+(defvar-keymap mac-window-menu-map :name "Window")
+(bindings--define-key global-map [menu-bar window]
+  (cons "Window" mac-window-menu-map))
+
 
 ;;
 ;; Standard Mac cursor shapes
@@ -3087,22 +3093,29 @@ standard ones in `x-handle-args'."
                     :button (:toggle . (mac-frame-tab-group-property
                                         nil :overview-visible-p)))
         'mac-toggle-tab-bar))
-    (define-key-after global-buffers-menu-map [mac-separator-tab]
+    (define-key-after mac-window-menu-map [mac-separator-tab]
       menu-bar-separator)
-    (define-key-after global-buffers-menu-map [mac-next-tab]
+    (define-key-after mac-window-menu-map [mac-next-tab]
       '(menu-item "Show Next Tab" mac-next-tab-or-toggle-tab-bar
                   :enable (mac-frame-multiple-tabs-p)))
     (global-set-key [(control tab)] 'mac-next-tab-or-toggle-tab-bar)
-    (define-key-after global-buffers-menu-map [mac-previous-tab]
+    (define-key-after mac-window-menu-map [mac-previous-tab]
       '(menu-item "Show Previous Tab" mac-previous-tab-or-toggle-tab-bar
                   :enable (mac-frame-multiple-tabs-p)))
     (global-set-key [(control shift tab)] 'mac-previous-tab-or-toggle-tab-bar)
-    (define-key-after global-buffers-menu-map [mac-move-tab-to-new-frame]
+    (define-key-after mac-window-menu-map [mac-move-tab-to-new-frame]
       '(menu-item "Move Tab to New Frame" mac-move-tab-to-new-frame
                   :enable (mac-frame-multiple-tabs-p)))
-    (define-key-after global-buffers-menu-map [mac-merge-all-frame-tabs]
+    (define-key-after mac-window-menu-map [mac-merge-all-frame-tabs]
       '(menu-item "Merge All Frames" mac-merge-all-frame-tabs
-                  :enable (mac-send-action 'mergeAllWindows t))))
+                  :enable (mac-send-action 'mergeAllWindows t)))
+
+    ;; In the macos menubar "Window" always comes before "Help"...
+    ;; Configure menu-bar-final-items to insert window before help.
+    (unless (memq 'window menu-bar-final-items)
+      (let* ((post-help (memq 'help-menu menu-bar-final-items))
+             (pre-help (butlast menu-bar-final-items (length post-help))))
+        (setq menu-bar-final-items (append pre-help '(window) post-help)))))
 
   (setq mac-ts-active-input-overlay (make-overlay 1 1))
   (overlay-put mac-ts-active-input-overlay 'display "")
--- a/lisp/term/mac-win.el
+++ b/lisp/term/mac-win.el
@@ -2882,6 +2882,11 @@ visibility, then remap this command to `mac-previous-tab'."
   (interactive)
   (mac-send-action 'mergeAllWindows))
 
+(defun mac-raise-all-frames ()
+  "Bring all frames (mac windows) to the front."
+  (interactive)
+  (mac-send-action 'arrangeInFront))
+
 
 ;;; Window system initialization.
 
@@ -3109,6 +3114,12 @@ standard ones in `x-handle-args'."
     (define-key-after mac-window-menu-map [mac-merge-all-frame-tabs]
       '(menu-item "Merge All Frames" mac-merge-all-frame-tabs
                   :enable (mac-send-action 'mergeAllWindows t)))
+
+    (define-key-after mac-window-menu-map [mac-separator-tab]
+      menu-bar-separator)
+    (define-key-after mac-window-menu-map [mac-merge-all-frame-tabs]
+      '(menu-item "Bring All to Front" mac-raise-all-frames
+                  :enable (mac-send-action 'arrangeInFront t)))
 
     ;; In the macos menubar "Window" always comes before "Help"...
     ;; Configure menu-bar-final-items to insert window before help.
--- a/src/macappkit.m
+++ b/src/macappkit.m
@@ -10647,7 +10647,7 @@ mac_font_dialog (struct frame *f)
 static void update_services_menu_types (void);
 static void mac_fake_menu_bar_click (EventPriority);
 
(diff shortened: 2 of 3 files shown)


jdtsmith:
Thanks for your contribution.  I'm a first time ObjC'er too, so no worries there.  I like the idea of a "Window" menu, though it may confuse Emacs users relative to Emacs' own definition of "windows".  I wonder, does it have to be called "Window"?  

For shortcuts, do you mean binding these in Settings' `App Shortcuts`?    *Update*: I see them under `Settings>Keyboard>Keyboard Shortcuts...>Windows`.  I use Hammerspoon for similar. 

Let me play around with this.

rymndhng:
@jdtsmith thanks, yeah, I agree that having a "Window" menubar could be confusing for emacs user. Some ideas...


### Option 1 -- Change the name

I can suggest "Tiling" or symbols that represent graphical tiling: 🪟, ⊞, 

Some screenshots:
Screenshot 2025-06-15 at 5 10 14 PM
Screenshot 2025-06-15 at 5 04 16 PM
Screenshot 2025-06-15 at 5 05 50 PM

### Option 2 -- Nest this menu under "Emacs"

I've put up a separate PR for this: https://github.com/jdtsmith/emacs-mac/pull/23

By nesting this under the "Emacs" menu item, this conveys the expectation that this is an apple-specific feature. Screenshot:

Screenshot 2025-06-15 at 5 39 00 PM

---

I personally would lean towards Option 2 to keep the menu bar tidy.

jdtsmith:
Thanks.  On the other hand, users of MacOS apps will absolutely expect and understand what a "Window" menu is for.  I just checked all my running apps, and every one has a "Window" menu item, so I think that's probably fine.

One question is about Tabs.  emacs-mac supports native tabs (which I use in preference to emacs tab-bar/tabline).  E.g. here is Safari's:

image

Did you come across any options for auto-enabling simple tab related commands?  We could simply move the commands from the bottom of the Buffers menu to a section in the new Window menu.

rymndhng:
Out of the box, I couldn't find anything that sets this up. Here are upstream docs for configuring menus, and all the menus are already configured in macappkit.m for mainMenu, servicesMenu, windowsMenu.

I played around a bit with tabs and it appears that emacs-mac treats each tab as a separate window. You get names of tabs for free ... but you do not get any controls for changing tabs.

Screenshot 2025-06-16 at 11 10 12 PM


I looked at a few apps (Firefox, Signal). I believe the apps with extra Window items customize this explicitly. For example, here's the [code](https://github.com/kovidgoyal/kitty/blob/4a13c534386ec97a33ae5570fdffe652db1b65c2/kitty/cocoa_window.m#L723-L780
) for kitty that produces the menu.

Screenshot: 
Screenshot 2025-06-16 at 11 07 33 PM

jdtsmith:
If you look at the bottom of the "Buffers" menu you'll see some hand-created Tab commands.  Maybe we can just move those over to the new "Window" menu. 

image

rymndhng:
@jdtsmith yeah, I noticed that. The configuration for these four commands are encoded in elisp, see https://github.com/jdtsmith/emacs-mac/blob/2226fd6ab970880285d5f75ec366d0febd7c0a4a/lisp/term/mac-win.el#L3092-L3105.

I think this is doable, however I'd tackle it differently from what I have now. I will try the following:

1. Create a new global menu map similar to `global-buffers-menu-map`, named `global-windows-menu-map`.
2. Assign these four commands to that window map.
3. In `mapappkit.m`, add code to find the "Window" menu, and call setWindowMenu on that menu. This is similar to how the existing code detects "Help".

I can give this a try

jdtsmith:
Excellent, thank you for looking into this.  Also, if you discover things about the interface between system-provided menus and elisp-defined menus (like how that works in general), it would be great to take some notes about it in the `devel_update_notes.org` file.

rymndhng:
@jdtsmith ready for another round of review.

Here's a screenshot of this the new Window menu with this patch:


Screenshot 2025-06-25 at 12 13 48 AM

lisp/term/mac-win.el
```diff
@@ -3087,22 +3093,30 @@ standard ones in `x-handle-args'."
...
                   :enable (mac-frame-multiple-tabs-p)))
-    (define-key-after global-buffers-menu-map [mac-merge-all-frame-tabs]
+    (define-key-after mac-window-menu-map [mac-merge-all-frame-tabs]
       '(menu-item "Merge All Frames" mac-merge-all-frame-tabs
-                  :enable (mac-send-action 'mergeAllWindows t))))
+                  :enable (mac-send-action 'mergeAllWindows t)))
+
+    ;; In the macos menubar "Window" always comes before "Help"...
+    ;; Configure menu-bar-final-items to insert window before help.
+    (let ((help-pos (or (cl-position 'help-menu menu-bar-final-items)
```
jdtsmith: I'd prefer not to have to load `cl`, since this file gets loaded right away.  Can you rewrite this with basic functions?  I'd suggest simply let-binding a nil list var, accumulate (via cons) all the cells up to "Help", then cons in "Help", then `setcdr` to the remaining tail and update the list.

src/macappkit.m
```diff
@@ -11330,6 +11336,11 @@ But first we recompute the menu bar contents (the whole tree).
 	  else if (title == localizedMenuTitleForHelp)
 	    helpMenu = submenu;
 
+	  /* Find the window menu for adding windows management items */
+	  if ([title isEqualToString:@"Window"])
+	    // TODO: Do we need to localize the title?
```
jdtsmith: I just tried and at least `Help` gets localized  (to "Hilfe" in German) currently.  I don't know where these localization strings are defined, perhaps there are system wide defaults for common menu names.  Why don't we just try it as it is setup for Help and Edit, and see if it works automatically in other languages (which I set in System Prefs>Language and Region).

jdtsmith (changes requested):
This is looking good, thanks, just a few things to look at.  I notice many Window menus have an item "Bring All to Front".  That would be good to have too, if it's straightforward.  Thanks again for your work on this.

jdtsmith:
Related question.  Shouldn't this be gated by `HAVE_MACGUI` too?

https://github.com/jdtsmith/emacs-mac/blob/2ef2f75d24b30b7214c426bcf2fb6c55b06d66dc/src/thread.c#L35-L39

This is relevant for `really_call_select`, which is what calls `mac_select`.  *Update*: wrong, in fact the `thread_select` call gets _replaced_, here:

https://github.com/jdtsmith/emacs-mac/blob/30901a29c103c019d4a68fcb48a3f2071b83b92d/src/xgselect.c#L188-L195

So nevermind.

what-the-functor:
This is really cool!

jdtsmith:
@rymndhng any progress?  Would be great to land this and then port to our master-branch tracking build.

rymndhng:
Been a bit busy recently. I’ll try to get back to it sometime this weekend.

I can rebase this off the base branch before applying your feedback.

jdtsmith:
No worries, thanks for working on it. Shouldn't be any changes to worry about.

lisp/term/mac-win.el
rymndhng: I ended up using c functions instead of cl. Does this work? https://github.com/jdtsmith/emacs-mac/pull/21/commits/b2c6fd3718e1d71527adbf8cab40ad7d3a68d9cb

src/macappkit.m
rymndhng: This was quite a rabbit hole... 

I followed these instructions [link](https://douglashill.co/localisation-using-apples-glossaries/), and it was not productive. I downloaded French, and was unable to find the localized strings.

Then, I asked Claude this question...

> On AppKit, show me how to get the localized string for the text "Window". I'll call 
> this in the function NSLocalizedStringFromTableInBundle

and to my amusement, it gave me the answer to the table I was looking for: `MenuCommands`.

rymndhng:
@jdtsmith This is ready for review again. I also fixed up a bug with localizing the Edit  menu.

On a mac, you can test the localization on individual apps under "Languages & Region > Applications". 

Here's a screenshot of Emacs Menubar in German ... also see that "Edit" is translated now.

Screenshot 2025-07-09 at 11 55 08 PM

rymndhng:
> This is looking good, thanks, just a few things to look at. I notice many Window menus have an item "Bring All to Front". That would be good to have too, if it's straightforward. Thanks again for your work on this.

I haven't figured out what enables this. I see it on many apps, but I also don't see it universally -- I don't see this on Firefox.

lisp/term/mac-win.el
jdtsmith: looks good.

jdtsmith:
I've tested this and it works great.  I also figured out there's a responder action called `arrangeInFront` that raises windows, so it was easy to add a menu item for that; see 65eace4.  (Well, I discovered this after I had implemented a full command in `macfns.c` and method in `macappkit.m` — good learning exercise).  

If you give this a test and it works for you, I think we're ready to merge.  Could you please squash all your commits but mine into one with a good commit message, and force push to this PR?  Thanks again for working on this.

rymndhng:
@jdtsmith looks good! I've squashed it down.

jdtsmith:
Thanks again for your work on this!  @simmsb would be good to merge this into the master tracking branch too.

jdtsmith:
Hi @rymndhng.  In case you missed it, users on older MacOS versions are encountering a crash due to the new `Window` menu (#64).  Please take a look if you have a moment.

rymndhng:
@jdtsmith thanks for the ping, I’ll take a look at it sometime this week. I have an older Ventura mac to test on.

jdtsmith:
@rymndhng did you see #140?  Let me know if you experience the same issue, and if you have a theory as to what could be causing this intermittent Fn-key handling.

PR #23, opened 2025-06-16
Add Window menu to appleMenu

rymndhng:
Nests the "Window" submenu under "Emacs".

Alternative to #21 

Example screenshot:

Screenshot 2025-06-15 at 5 39 00 PM


jdtsmith:
I think we can stick with the `Window` menu as in #21.

Issue #43, opened 2025-07-25
homebrew Clang leads to EXC_BAD_ACCESS


jdtsmith:
See #41.  Typical output:

```
(lldb) run
Process 94411 launched: '/tmp/Emacs.app/Contents/MacOS/Emacs' (arm64)
2025-07-24 18:55:30.553260-0400 Emacs[94411:39014338] [logging-persist] cannot open file at line 49448 of [1b37c146ee]
2025-07-24 18:55:30.553277-0400 Emacs[94411:39014338] [logging-persist] os_unix.c:49448: (2) open(/private/var/db/DetachedSignatures) - No such file or directory
Process 94411 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x2d1097f70038)
    frame #0: 0x000000018bc1c028 libobjc.A.dylib`objc_release + 16
libobjc.A.dylib`objc_release:
->  0x18bc1c028 <+16>: ldr    x17, [x2, #0x20]
    0x18bc1c02c <+20>: tbz    w17, #0x2, 0x18bc1c090 ; <+120>
    0x18bc1c030 <+24>: tbz    w16, #0x0, 0x18bc1c0b4 ; <+156>
    0x18bc1c034 <+28>: lsr    x17, x16, #55
(lldb) up 3
frame #3: 0x0000000100443d44 Emacs`mac_gui_loop at macappkit.m:16617:7
   16614	      if (block)
   16615		block ();
   16616	      dispatch_semaphore_signal (mac_lisp_semaphore);
-> 16617	      END_AUTORELEASE_POOL;
   16618	    }
   16619	  while (block);
   16620	}
(lldb) p __clang_version__
(const char[8]) "20.1.8 "
(lldb) p __clang_major__
(int) 20
```

This can be fixed by bringing the AUTORELEASE directly around `block()`.  Then pulling up the font selection pane leads to another.  One thought is that Clang v20 is much more aggressive and strict about memory pool access, so is revealing latent memory release issues.

gerd-moellmann:
Tried with Homebrew clang 17, and that also doesn't work, like with clang 20. 

In the Console app, I see in the crash report:

```
Thread 0::  Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	       0x1915ee978 __sigsuspend + 8
[...]

PR #48, opened 2025-07-27
Do not use CF_NOESCAPE

jdtsmith:
CF_NOESCAPE is __attribute__ ((noescape)), and there is at least
one place where the promise that the block does not escape is
not true because it is stored in a queue. This leads to
undefined behavior when compiled with non-Apple clang.

* src/mac.c (mac_foreach_window_1):
(mac_foreach_window):
(mac_map_keymap):
* src/macappkit.m (NS_TOUCH_BAR_ITEM_IDENTIFIER_CANDIDATE_LIST):
(mac_autorelease_loop):
(mac_within_gui):
(mac_within_gui_and_here):
(mac_within_gui_allowing_inner_lisp):
(mac_within_lisp):
* src/macgui.h (CF_NOESCAPE):
* src/macterm.h:
* src/nsxwidget.m (mac_within_gui_check_thread):

### DIFF
--- a/src/mac.c
+++ b/src/mac.c
@@ -138,7 +138,7 @@ mac_string_to_four_char_code (Lisp_Object string, FourCharCode *code)
 
 static bool
 mac_foreach_window_1 (struct window *w,
-		      bool (CF_NOESCAPE ^block) (struct window *))
+		      bool (^block) (struct window *))
 {
   bool cont;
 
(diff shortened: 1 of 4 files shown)


jdtsmith:
@simmsb, I decided to break (refine) my rule with this one and bring specific changes to the emacs-mac side direct to your master tracking branch.  I think this could use a bit more testing.  See #43 for the details (apparently Apple CLANG is subtly different that upstream CLANG for the same version, w.r.t. blocks and the compiler hint in CF_NOESCAPE).

jdtsmith:
I'll bravely go ahead and merge this; please let me know if you encounter any memory issues.

PR #61, opened 2025-07-30
Make self-contained build the default

jdtsmith:
* carbon/README: Added.
* configure.ac: Replace --enable-mac-self-contained with
--disable-mac-self-contained. Use a default installation
directory of ./carbon. Remove --enable-mac-app=yes to mean
installation in /Applications.\

See #58 for some discussion.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -667,16 +667,16 @@ case ${with_gameuser} in
   *) gameuser=${with_gameuser} ;;
 esac
 
-AC_ARG_ENABLE(mac-app,
+AC_ARG_ENABLE([mac-app],
 [AS_HELP_STRING([--enable-mac-app@<:@=DIR@:>@],
-                [specify install directory for Emacs.app on macOS
-		 [DIR=/Application]])],
-[ mac_appdir_x=${enableval}])
+                [specify install directory for Emacs.app on macOS.])],
+   [mac_appdir_x=${enableval}])
 
-AC_ARG_ENABLE(mac-self-contained,
-[AS_HELP_STRING([--enable-mac-self-contained],
-                [enable self contained build under mac])],
-                EN_MAC_SELF_CONTAINED=$enableval)
+AC_ARG_ENABLE([mac-self-contained],
+[AS_HELP_STRING([--disable-mac-self-contained],
+                [disable self contained build under mac])],
+   [EN_MAC_SELF_CONTAINED=$enableval],
+   [EN_MAC_SELF_CONTAINED=yes])
 
 AC_ARG_WITH([gnustep-conf],
 [AS_HELP_STRING([--with-gnustep-conf=FILENAME],
@@ -2866,12 +2866,9 @@ if test "${with_mac}" != no; then
   test "${HAVE_MACGUI}" = yes && window_system=mac
 
   ## Specify the install directory
-  mac_appdir=`pwd`/mac
+  mac_appdir=`pwd`/carbon
   if test "${mac_appdir_x}" != ""; then
-    case ${mac_appdir_x} in
-      y | ye | yes)  mac_appdir=/Applications ;;
-      * ) mac_appdir=${mac_appdir_x} ;;
-    esac
+    mac_appdir=${mac_appdir_x}
   fi
 
   mac_appbindir=${mac_appdir}/Emacs.app/Contents/MacOS
--- a/carbon/README
+++ b/carbon/README
@@ -1,0 +1,2 @@
+This directory contains the self-contained Emacs.app application.
+You can copy this to where you want, e.e. /Applications.


configure.ac
```diff
@@ -2866,12 +2866,9 @@ if test "${with_mac}" != no; then
   test "${HAVE_MACGUI}" = yes && window_system=mac
 
   ## Specify the install directory
-  mac_appdir=`pwd`/mac
+  mac_appdir=`pwd`/carbon
```
jdtsmith, line 2869: Given how confusing "Carbon" is, and that it's an API discontinued >10yrs ago, I think we may want to phase out that brand over time.  What's wrong with reusing `./mac/`?

jdtsmith:
@gerd-moellmann This still requires `make install` to get a functioning app bundle, yes?

jdtsmith:
From #58, @gerd-moellmann said:
> Please scratch that. On second thought, and after having looked at the
Makefiles, it's less work to Add a carbon directory to the repo, i.e.
add a new file carbon/README, and install there by default, i.e.
--enable-mac-app=./carbon.

I see.  But then I don't understand the point as much.  `make install` is still required, and the "intermediate stop" of `mac/Emacs.app` which exists after `make` is not a complete bundle.  I think it's an improvement to have this on by default, but how does it service the goal of not requiring "make install" so often?   

Put another way, why is this better than defaulting to `mac-self-contained=yes`, but continuing to  target `/Applications' for install?

jdtsmith:
> And the below could be the reason why mac/ sometimes ends up being inconsistent, IMO:
> configure.ac<emacs-mac>:
> ``` 2793   ## Specify the install directory
> 2794   mac_appdir=`pwd`/mac
> 2795   if test "${mac_appdir_x}" != ""; then
> ```
> If mac_appdir stays to be mac, mac/Emacs.app is both copied and overwritten when installing. Not good.

I agree this is pretty unhygienic, and have considered that's what leads to some misfeatures like outdated pdmp etc..  But why not create a `mac/Emacs.app.template`, which can be copied and filled out under `mac/Emacs.app`?

gerd-moellmann:
JD Smith ***@***.***> writes:

> @jdtsmith commented on this pull request.
>
> -------------------------------------------------------------------------------------------------------------------------
>
> In configure.ac:
>
>> @@ -2866,12 +2866,9 @@ if test "${with_mac}" != no; then
>    test "${HAVE_MACGUI}" = yes && window_system=mac
>  
>    ## Specify the install directory
> -  mac_appdir=`pwd`/mac
> +  mac_appdir=`pwd`/carbon
>
> Given how confusing "Carbon" is, and that it's an API discontinued >10yrs ago, I think we may want to phase out that brand over
> time. What's wrong with reusing ./mac/?

I'd prefer the name mac, too, compare my priginal proposal, but it is
more work that I don't wang go invest ATM.

The problem is that a mac/Emacs.app exists. This is copied to the final
Emacs.app during installation. Setting mac_appdir to mac means that
mac/Emacs.app is copied to itself, so to speak. See mac/Makefile.in.

gerd-moellmann:
JD Smith ***@***.***> writes:

> *
> jdtsmith left a comment (jdtsmith/emacs-mac#61)
>
> @gerd-moellmann This still requires make install to get a functioning
> app bundle, yes?

Correct.

gerd-moellmann:
JD Smith ***@***.***> writes:

> *
> jdtsmith left a comment (jdtsmith/emacs-mac#61)
>
> From #58, @gerd-moellmann said:
>
>  Please scratch that. On second thought, and after having looked at the
>  Makefiles, it's less work to Add a carbon directory to the repo, i.e.
>  add a new file carbon/README, and install there by default, i.e.
>  --enable-mac-app=./carbon.
>
> I see. But then I don't understand the point as much. make install is
> still required, and the "intermediate stop" of mac/Emacs.app which
> exists after make is not a complete bundle. I think it's an
> improvement to have this on by default, but how does it service the
> goal of not requiring "make install" so often?

Not requiring a make install to be able to start Emacs was not my goal
with this patch. That goal is served with the patch I sent for starting
src/emacs.

The goal of this patch is to improve the configuration and installation
process so that mac and ns are built basically in the same way
(self-contained, and in a subdirectory of the repp).

gerd-moellmann:
JD Smith ***@***.***> writes:

> jdtsmith left a comment (jdtsmith/emacs-mac#61)
>
>  And the below could be the reason why mac/ sometimes ends up being inconsistent, IMO:
>  configure.ac:
>
> 2794   mac_appdir=`pwd`/mac
> 2795   if test "${mac_appdir_x}" != ""; then
>
>  If mac_appdir stays to be mac, mac/Emacs.app is both copied and overwritten when installing. Not good.
>
> I agree this is pretty unhygienic, and have considered that's what
> leads to some misfeatures like outdated pdmp etc.. But why not create
> a mac/Emacs.app.template, which can be copied and filled out under
> mac/Emacs.app?

That's basically wanted to do originally, rename the mac/Emacs.app to
mac/Emacs.app.template, or something. It's more work because the
Makefiles also have to be changed. Didn't want to invest that time
simply for a nicer name :-)

jdtsmith:
So in your approach, the user themselves has to copy `carbon/Emacs.app`  into `/Applications/`? That's a pretty big change.  Is that how NS works? (I have compiled it only rarely.)

I presume your build scripts could instead just `--enable-mac-app=./carbon`, even if we make `enable-mac-app=/Applications` the default.  I'll think about it. @simmsb, any thoughts on the matter?

gerd-moellmann:
JD Smith ***@***.***> writes:

> *
> jdtsmith left a comment (jdtsmith/emacs-mac#61)
>
> So in your approach, the user themselves has to copy carbon/Emacs.app
> into /Applications/? That's a pretty big change. Is that how NS works?
> (I have compiled it only rarely.)

Yes, that's how NS works.

> I presume your build scripts could instead just
> --enable-mac-app=./carbon, even if we make
> enable-mac-app=/Applications the default. I'll think about it.
> @simmsb, any thoughts on the matter?

That's what my make-emacs does, when I specify --mac plus
--self-contained, because it it doesn't have to in my branch :-). (I
have to say --mac, or --ns, because I normally build without
window-system.)

simmsb:
>  @simmsb, any thoughts on the matter?
Self contained is definitely the better default, I also would prefer the default to be placing the .app somewhere in the source folder.

jdtsmith:
OK, let's wait on this for a bit and see if any opinions emerge upstream.  I guess most current users are getting it from home-brew, which would adapt to the new compile/install flow anyway.  The `nextstep/INSTALL` file simply says:

> This will assemble the app in nextstep/Emacs.app (i.e., the --prefix argument has no effect in this case).  You can then move the Emacs.app bundle to a location of your choice.

Issue #69, opened 2025-08-02
`NS_NOESCAPE`


gerd-moellmann:
There are declarations in the mac code that use `NS_NOESCAPE` which is defined to be `CF_NOESCAPE` in some framework header. For example, 
```
static void
mac_with_current_drawing_appearance (NSAppearance *appearance,
				     void (NS_NOESCAPE ^block) (void))
```
This is potentially as dangerous with non-system clang as using `CF_NOESCAPE` was.  I have removed the uses of `NS_NOESCAPE?  in my Emacs.

[0001-Do-not-use-NS_NOESCAPE.patch](https://github.com/user-attachments/files/21557571/0001-Do-not-use-NS_NOESCAPE.patch)

jdtsmith:
Do we _know_ that `NS/CF_NOESCAPE` is a real problem in all these cases?  I'm only barely familiar, but I think noescape is generally useful for both compiler optimization and avoiding memory leaks, when the block gets executed directly in the method to which it is passed.  Since we make and pass around so many blocks, it makes sense to keep these if they do no harm.  

@tnytown do you have an opinion?

gerd-moellmann:
JD Smith ***@***.***> writes:

> *
> jdtsmith left a comment (jdtsmith/emacs-mac#69)
>
> Do we know that NS/CF_NOESCAPE is a real problem in all these cases?
> I'm only barely familiar, but I think noescape is generally useful for
> both compiler optimization and avoiding memory leaks, when the block
> gets executed directly in the method to which it is passed. Since we
> make and pass around so many blocks, it makes sense to keep these if
> they do no harm.

CF_NOESCAPE is a promise made to the compiler. If the promise is false,
one has UB. In the case of the use of CF_NOESCAPE in the mac sources,
the noescape was a lie, the blocks escaped. I showed an example in the
ticket.

I couldn't be bothered to check the uses of NS_NOESCAPE, arguing that
after the first lie, the burden of proof lies on someone wanting to use
it :-).

tnytown:
> Do we know that NS/CF_NOESCAPE is a real problem in all these cases? I'm only barely familiar, but I think noescape is generally useful for both compiler optimization and avoiding memory leaks, when the block gets executed directly in the method to which it is passed. Since we make and pass around so many blocks, it makes sense to keep these if they do no harm.

[...]

PR #76, opened 2025-08-10
fix: Remove excessive CF_NOESCAPE and NS_NOESCAPE

pkryger:
Analysed all occurrences of CF_NOESCAPE and NS_NOESCAPE and removed
where the annotation seemed to be excessive, for example when the
annotated block has been enqueued in a queue to execute in another loop.

Such a block, may have a an optimised placement as
`NSConcreteGlobalBlock` on stack on a callee side. This happen on
non-Apple clang Apple at least in version 7 (see [1]).

This is slightly different than fully disabling the blocks (see [2],
[3], and [4]), while trying to preserve as much of the NOESCAPES where
it seems to server a purpose.

* src/macappkit.m (mac_within_gui, mac_within_lisp,
  mac_within_gui_and_here, mac_within_gui_allowing_inner_lisp): Remove
  excessive CF_NOESCAPE, for blocks that are passed and executed in
  other threads.
* src/macterm.h (mac_within_gui) : Remove excessive CF_NOESCAPE, for
  blocks that are passed and executed in other threads.
* src/macappkit.h (enumerateChildWindowsUsingBlock): Harmonise
  NS_NOESCAPE placement.
* src/macappkit.m (enumerateChildWindowsUsingBlock): Harmonise
  CF_NOESCAPE placement.

[1] https://github.com/NixOS/nixpkgs/issues/127902#issuecomment-1698365843
[2] https://github.com/jdtsmith/emacs-mac/issues/43
[3] https://github.com/jdtsmith/emacs-mac/pull/48
[4] https://github.com/jdtsmith/emacs-mac/issues/69

### DIFF
--- a/src/macappkit.h
+++ b/src/macappkit.h
@@ -508,8 +508,8 @@ typedef NSInteger NSGlyphProperty;
 @interface NSWindow (Emacs)
 - (Lisp_Object)lispFrame;
 - (NSWindow *)topLevelWindow;
-- (void)enumerateChildWindowsUsingBlock:(NS_NOESCAPE void
-					 (^)(NSWindow *child, BOOL *stop))block;
+- (void)enumerateChildWindowsUsingBlock:(void
+					 (NS_NOESCAPE ^)(NSWindow *child, BOOL *stop))block;
 @end
 
 @interface NSCursor (Emacs)
--- a/src/macappkit.m
+++ b/src/macappkit.m
@@ -109,10 +109,10 @@ enum {
   (@"NSTouchBarItemIdentifierCandidateList")
 #endif
 
-static void mac_within_gui_and_here (void (^ CF_NOESCAPE) (void),
-				     void (^ CF_NOESCAPE) (void));
-static void mac_within_gui_allowing_inner_lisp (void (^ CF_NOESCAPE) (void));
-static void mac_within_lisp (void (^ CF_NOESCAPE) (void));
+static void mac_within_gui_and_here (void (^) (void),
+                                     void (CF_NOESCAPE ^) (void));
+static void mac_within_gui_allowing_inner_lisp (void (^) (void));
+static void mac_within_lisp (void (^) (void));
 static void mac_within_lisp_deferred_unless_popup (void (^) (void));
 
 #define MAC_SELECT_ALLOW_LISP_EVALUATION 1
@@ -627,8 +627,8 @@ mac_within_app (void (^block) (void))
    descendant windows.  Set *stop to YES in the block to abort further
    processing of the child windows subtree.  */
 
-- (void)enumerateChildWindowsUsingBlock:(NS_NOESCAPE void
-					 (^)(NSWindow *child, BOOL *stop))block
+- (void)enumerateChildWindowsUsingBlock:(void
+					 (NS_NOESCAPE ^)(NSWindow *child, BOOL *stop))block
 {
   for (NSWindow *childWindow in self.childWindows)
     {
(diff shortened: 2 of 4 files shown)


src/macappkit.h
```diff
@@ -508,8 +508,8 @@ typedef NSInteger NSGlyphProperty;
 @interface NSWindow (Emacs)
 - (Lisp_Object)lispFrame;
 - (NSWindow *)topLevelWindow;
-- (void)enumerateChildWindowsUsingBlock:(NS_NOESCAPE void
-					 (^)(NSWindow *child, BOOL *stop))block;
+- (void)enumerateChildWindowsUsingBlock:(void
+					 (NS_NOESCAPE ^)(NSWindow *child, BOOL *stop))block;
```
jdtsmith, line 512: Is this just for consistency or does the order of the caret matter?

pkryger: That's right, this is for consistency only, the order doesn't matter.

<details>

In a context of [this gist](https://gist.github.com/pkryger/9a365e069173e5effc6f90e85e473fbf).

I compiled a small example (just most important bits):
```objective-c
#define MY_NOESCAPE __attribute__ ((noescape))
void queue_block_1(void (^ MY_NOESCAPE block) (void)) {...}
void queue_block_2(void (MY_NOESCAPE ^block) (void)) {...}
void queue_block_3(MY_NOESCAPE void (^block) (void)) {...}
void queue_1(const char* arg)
{
  queue_block_1(^{
      printf("block_1: %s\n", arg);
    });
}

void queue_2(const char* arg)
{
  queue_block_2(^{
      printf("block_2: %s\n", arg);
    });
}

void queue_3(const char* arg)
{
  queue_block_3(^{
      printf("block_3: %s\n", arg);
    });
}
```

with `$(brew --prefix llvm)/bin/clang -g -framework CoreFoundation -o reproducer_123  reproducer.m` and then:
```
(lldb) dis -n queue_1
reproducer_123`queue_1:
[...]
reproducer_123[0x100000828] <+32>: mov    w8, #-0x2f800000 ; =-796917760
[...]
(lldb) dis -n queue_2
reproducer_123`queue_2:
[...]
reproducer_123[0x1000008c4] <+32>: mov    w8, #-0x2f800000 ; =-796917760
(lldb) dis -n queue_3
reproducer_123`queue_3:
[...]
reproducer_123[0x100000960] <+32>: mov    w8, #-0x2f800000 ; =-796917760
```
</details>

src/macappkit.m
```diff
@@ -109,10 +109,10 @@
   (@"NSTouchBarItemIdentifierCandidateList")
 #endif
 
-static void mac_within_gui_and_here (void (^ CF_NOESCAPE) (void),
-				     void (^ CF_NOESCAPE) (void));
-static void mac_within_gui_allowing_inner_lisp (void (^ CF_NOESCAPE) (void));
-static void mac_within_lisp (void (^ CF_NOESCAPE) (void));
+static void mac_within_gui_and_here (void (^) (void),
+                                     void (CF_NOESCAPE ^) (void));
```
jdtsmith, line 113: Here the idea is that `block_here` is executed locally, and so does not escape?

pkryger: Exactly that.

[This](https://github.com/jdtsmith/emacs-mac/blob/87ded1c7386599083969ca7ba0172de295b7e045/src/macappkit.m#L16671-L16672) is the only usage of `block_here`

jdtsmith:
Thanks for your contribution!

PR #81, opened 2025-08-12
Support 'dots' and 'dashes' underline styles

woolsweater:
This adds support for the new face underline styles, `(:style dots)` and `(:style dashes)`.

Shown here applied as overlays, with the default color `(:underline (:style dots))` and with a custom color `(:underline (:style dots :color "magenta"))`.

Screenshot 2025-08-11 at 6 26 51 PM

### DIFF
--- a/src/macterm.c
+++ b/src/macterm.c
@@ -64,6 +64,13 @@ static Lisp_Object x_display_rdb_list;
 /* This is display since Mac does not support multiple ones.  */
 struct mac_display_info one_mac_display_info;
 
+/* Patterns for CGContextSetLineDash for dotted and dashed
+   underlines. The first number is the length of the drawn segment, the
+   second the length of the gap between.  */
+
+static const CGFloat mac_underdot_pattern[2] = { 1, 2 };
+static const CGFloat mac_underdash_pattern[2] = { 3, 2 };
+
 /* The keysyms to use for the various modifiers.  */
 
 static struct terminal *mac_create_terminal (struct mac_display_info *dpyinfo);
@@ -378,6 +385,33 @@ mac_draw_horizontal_wave (struct frame *f, GC gc, int x, int y,
 	gx1 += gperiod;
 	CGContextAddLineToPoint (context, gx1, gy1);
       }
+    CGContextStrokePath (context);
+  }
+  MAC_END_DRAW_TO_FRAME (f);
+}
+
+static void
+mac_draw_dashed_line (struct frame *f, GC gc, int x, int y, int width,
+		      CGFloat thickness, const CGFloat *pattern)
+{
+  CGRect clip = CGRectMake (x, y, width, thickness * 2);
+
+  MAC_BEGIN_DRAW_TO_FRAME (f, gc, clip, context);
+  {
+    CGFloat gxmax, gy, gpattern[2];
+
+    gxmax = CGRectGetMaxX (clip);
+    gy = y + (thickness / 2);
+    // Make the segments proportionate to thickness.
+    gpattern[0] = pattern[0] * thickness;
+    gpattern[1] = pattern[1] * thickness;
+
+    CGContextClipToRect (context, clip);
+    CGContextMoveToPoint (context, x, gy);
+    CGContextAddLineToPoint (context, gxmax, gy);
+    CGContextSetStrokeColorWithColor (context, gc->cg_fore_color);
+    CGContextSetLineWidth (context, thickness);
+    CGContextSetLineDash (context, 0, gpattern, 2);
     CGContextStrokePath (context);
   }
   MAC_END_DRAW_TO_FRAME (f);
@@ -2240,6 +2274,13 @@ mac_draw_underwave (struct glyph_string *s, int decoration_width)
 			    decoration_width, wave_height, wave_length);
 }
 
+static void
+mac_draw_underdash (struct glyph_string *s, int decoration_width, int y,
+		    CGFloat thickness, const CGFloat *pattern)
+{
+    mac_draw_dashed_line (s->f, s->gc, s->x, y, decoration_width,
+			  thickness, pattern);
+}
 
 /* Draw glyph string S.  */
 
@@ -2379,10 +2420,14 @@ mac_draw_glyph_string (struct glyph_string *s)
 		  mac_set_foreground (s->gc, xgcv.foreground);
 		}
 	    }
-	  else if (s->face->underline == FACE_UNDERLINE_SINGLE)
+	  else if (s->face->underline == FACE_UNDERLINE_SINGLE
+		   || s->face->underline == FACE_UNDERLINE_DOTS
+		   || s->face->underline == FACE_UNDERLINE_DASHES)
 	    {
 	      unsigned long thickness, position;
 	      int y;
+	      bool did_set_foreground = false;
+	      XGCValues xgcv;
 
               if (s->prev
 		  && s->prev->face->underline == FACE_UNDERLINE_SINGLE
@@ -2464,16 +2509,27 @@ mac_draw_glyph_string (struct glyph_string *s)
 	      s->underline_thickness = thickness;
 	      s->underline_position = position;
 	      y = s->ybase + position;
-	      if (s->face->underline_defaulted_p)
+
+	      if (!s->face->underline_defaulted_p)
+		{
+		  mac_get_gc_values (s->gc, GCForeground, &xgcv);
+		  mac_set_foreground (s->gc, s->face->underline_color);
+		  did_set_foreground = true;
+		}
+
+	      if (s->face->underline == FACE_UNDERLINE_SINGLE)
 		mac_fill_rectangle (s->f, s->gc, s->x, y, s->width, thickness);
 	      else
 		{
-		  XGCValues xgcv;
-		  mac_get_gc_values (s->gc, GCForeground, &xgcv);
-		  mac_set_foreground (s->gc, s->face->underline_color);
-		  mac_fill_rectangle (s->f, s->gc, s->x, y, s->width, thickness);
-		  mac_set_foreground (s->gc, xgcv.foreground);
+		  const CGFloat *pattern =
+		    (s->face->underline == FACE_UNDERLINE_DOTS)
+		    ? mac_underdot_pattern : mac_underdash_pattern;
+		  mac_draw_underdash (s, decoration_width, y,
+				      thickness, pattern);
 		}
+
+	      if (did_set_foreground)
+		mac_set_foreground (s->gc, xgcv.foreground);
 	    }
 	}
 
--- a/src/macterm.c
+++ b/src/macterm.c
@@ -2282,6 +2282,133 @@ mac_draw_underdash (struct glyph_string *s, int decoration_width, int y,
 			  thickness, pattern);
 }
 
+static void
+mac_draw_underline (struct glyph_string *s, int decoration_width)
+{
+  unsigned long thickness, position;
+  int y;
+  bool did_set_foreground = false;
+  XGCValues xgcv;
+  const CGFloat *pattern = mac_underdot_pattern;
+
+  if (s->prev
+      && s->prev->face->underline == FACE_UNDERLINE_SINGLE
+      && (s->prev->face->underline_at_descent_line_p
+	  == s->face->underline_at_descent_line_p)
+      && (s->prev->face->underline_pixels_above_descent_line
+	  == s->face->underline_pixels_above_descent_line))
+    {
+      /* We use the same underline style as the previous one.  */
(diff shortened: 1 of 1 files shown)


src/macterm.c
```diff
@@ -64,6 +64,13 @@ static Lisp_Object x_display_rdb_list;
 /* This is display since Mac does not support multiple ones.  */
 struct mac_display_info one_mac_display_info;
 
+/* Patterns for CGContextSetLineDash for dotted and dashed
+   underlines. The first number is the length of the drawn segment, the
+   second the length of the gap between.  */
+
+static const CGFloat mac_underdot_pattern[2] = { 1, 2 };
+static const CGFloat mac_underdash_pattern[2] = { 3, 2 };
```
woolsweater, line 72: ℹ️ Defined these statically because with `DRAWING_USE_GCD`, `MAC_BEGIN_DRAW_TO_FRAME` is asynchronous and an array created on the stack and passed in to `mac_draw_dashed_line` becomes garbage. (There are other options but this seemed like the most straightforward.)

src/macterm.c
```diff
@@ -2379,10 +2420,14 @@ mac_draw_glyph_string (struct glyph_string *s)
 		  mac_set_foreground (s->gc, xgcv.foreground);
 		}
 	    }
-	  else if (s->face->underline == FACE_UNDERLINE_SINGLE)
+	  else if (s->face->underline == FACE_UNDERLINE_SINGLE
```
woolsweater: ℹ️ I reused this branch for the thickness and position calculations, which I think we want for these styles as we do for a plain underline.

woolsweater:
Sorry, withdrawing this after seeing https://github.com/jdtsmith/emacs-mac/discussions/71

jdtsmith:
Thanks. Does that mean you don't yet have copyright assignment on file, or don't want to assign?  For now this is hypothetical, since the prospects for upstreaming are not clear.

woolsweater:
> Thanks. Does that mean you don't yet have copyright assignment on file, or don't want to assign? For now this is hypothetical, since the prospects for upstreaming are not clear.

The latter; without wanting to get too deep here 😅 , I'm not really on board with (the FSF's position on) copyright assignment. In my opinion, if I'm contributing to a GPL project, the GPL should be sufficient for my contribution as well (the "inbound=outbound" principle).

I would be very happy to contribute to this fork using the GPL, if that is sufficient.

jdtsmith:
I understand.  It's possible this point is moot, since the majority of unique code here (many 10s of thousands of lines) is from Yamamoto-san , who has not assigned copyright for it.  If it looks like that situation will not change, we may drop the requirement.

pkryger:
@jdtsmith Have you made a decision whether or not presence of GPL copyright assignment is a blocker? I think I found a use case for this feature.

jdtsmith:
I'd like to preserve the option for copyright assignment for a bit longer, in case an interest materializes with the next release of Emacs (unlikely, but you never know).

As for this feature, it seems like it's fairly simple functionality.  Perhaps you could reimplement it yourself (ideally without looking in detail at the code)?

BTW, is there a reason not to propose such a feature upstream (via NS)?  Obviously it would be more useful if 3rd party packages could reliably target dot/dash style.

woolsweater:
> BTW, is there a reason not to propose such a feature upstream (via NS)? Obviously it would be more useful if 3rd party packages could reliably target dot/dash style.

It's already implemented for NS: https://github.com/jdtsmith/emacs-mac/blob/96927f1f70fa05a23a832defb87af20a43eecf4b/src/nsterm.m#L3309-L3368 The implementation just isn't directly reusable for the Mac Port since macterm.c uses CoreGraphics rather than Cocoa for drawing.

jdtsmith:
Oh, interesting, missed that.  I see that `:style` is described as:

>  one of ‘line’, ‘double-line’, ‘wave’, ‘dots’, or ‘dashes’.

Do we support `double-line` already?

woolsweater:
Looks like NS supports it but Mac ignores the "double" part and just draws a single line.

jdtsmith:
I'm going to reopen this to keep it on the radar.  Might be nice to add "double" support for completeness.  The fact that this is already in other builds is a strong mark in favor.

woolsweater:
> Might be nice to add "double" support for completeness.

Yeah, definitely, I didn't realize before that it was missing. I'll add it; I should be able to get to it sometime in the next week. Although it feels a little silly if this isn't going to be merged because of licensing 😄

jdtsmith:
Thanks.  A patch is useful even if it has to be maintained in one of the downstream package repositories.  That said, I suspect this will eventually be merged.

woolsweater:
> A patch is useful even if it has to be maintained in one of the downstream package repositories.

Fair poiint! 👍

pkryger:
> > Might be nice to add "double" support for completeness.
> 
> Yeah, definitely, I didn't realize before that it was missing. I'll add it; I should be able to get to it sometime in the next week. Although it feels a little silly if this isn't going to be merged because of licensing 😄 

Thanks for giving it another go. I think something needs to be done for underdouble (is this the term 🤔). Reading the code it gives me impression it would draw underdashes when `double-line` is requested.

FWIW what people use for a terminal? It seems that on iTerm2, even when terminal supports `Smulx` capability only underline and underwave are suppoeretd (falling back to former for other styles). I have a prototype patch where it can be made to draw underdouble and underdashes (imperfectly). But no underdots. My current understanding is that it's more iTerm2 issue rather than Emacs, but wanted to ask here first.

pkryger:
I thing I managed to convince myself support for all underline styles is a terminal responsibility. Emacs is capable of rendering all supported underline styles in non-GUI frame when terminal supports all of them via `Smulx` capability. For example `kitty` renders all of them:
Screenshot 2025-12-02 at 16 40 38

I guess that's one more reason to add support in a GUI frame.

I have logged [an issue for `iTerm2`](https://gitlab.com/gnachman/iterm2/-/issues/12633) to add support for it.

woolsweater:
Okay, here we are:

Screenshot 2025-12-11 at 8 06 04 PM

As noted in the commit message, the positioning of the second line is the same as in NS, Haiku, and probably others, but it sure does look ugly to me 😅 Maybe we can refine it in the future to look better though.

src/macterm.c
```diff
@@ -2240,6 +2274,140 @@ mac_draw_underwave (struct glyph_string *s, int decoration_width)
...
+  y = s->ybase + position;
+
+  if (!s->face->underline_defaulted_p)
+    {
+      mac_get_gc_values (s->gc, GCForeground, &xgcv);
+      mac_set_foreground (s->gc, s->face->underline_color);
+      did_set_foreground = true;
+    }
+
+  switch (s->face->underline)
```
woolsweater, line 2383: ℹ️ This switch is really the only new bit here; everything else just moved from the `else if` in `mac_draw_glyph_string` below.

woolsweater:
> support for all underline styles is a terminal responsibility

It is. Another thing I would like to change about the Mac Port at some point is to make the underlines drawn _behind_ the characters. You can see in my original screenshot here with the colored underlines that they are in front.

pkryger:
> > support for all underline styles is a terminal responsibility
> 
> 
> 
> It is.

Thanks for confirming. iTerm2 has recently added [support](https://gitlab.com/gnachman/iterm2/-/commit/f98ef284e94420ee23bf46e5f31c0cf0fc66524d) (waiting for a beta build to test it), while `kitty` has the support implemented for some time. So only GUI of Emacs-mac is lacking it 🫣.

> Another thing I would like to change about the Mac Port at some point is to make the underlines drawn _behind_ the characters. You can see in my original screenshot here with the colored underlines that they are in front.

That would be cool. My experience in GUI programming is almost non-existent, so I don't know how practical it would be but an alternative could be to add gaps in the underlines. Like iTerm2 and `kitt` do (see [screenshots](https://gitlab.com/gnachman/iterm2/-/issues/12633).

jdtsmith:
Thanks.  Is it possible to reduce line thickness and move the top line of the pair down somewhat?  Can you see a simple way to reverse the order of the drawing of glyphs and lines?  Our underlines don't look exactly the same as NS's, and that's fine.

src/macterm.c
jdtsmith: I wonder if you can move all the under/overline code ahead of the glyph drawing?  Not sure if that would interfere with filled background.

woolsweater: Yes, I think so, but I'd prefer to do that as its own patch so we keep each change focused.

woolsweater:
> Is it possible to reduce line thickness and move the top line of the pair down somewhat?

Maybe, but it's a bigger change. The lines are actually drawn as rectangles where the thickness is used as the rectangle height, which is an integer. (I'm not sure why it's done that way.) The line thickness is generally going to be 1 point (it defaults to that but the font definition also has a say), so reducing it might run into antialiasing problems. Additionally moving the top line down might mean needing to adjust the lower line as well, which means we are ignoring the font's definition of where its underline goes, which probably will look bad in its own way in some cases.

I'm willing to fiddle around with it, but I won't be able to get to that for a little while.

woolsweater:
> an alternative could be to add gaps in the underlines.

Yeah, I do think that looks pretty good. TextEdit does the same:

Screenshot 2025-12-14 at 6 08 06 PM

but it involves typography beyond my knowledge since it requires looking at the metrics of each character. This is probably something that AppKit or CoreText handles "automatically" but I don't think that can be used for Emacs because it would mean giving up layout control.

Interestingly, Pages draws the underline on _top_ of the characters:

Screenshot 2025-12-14 at 6 10 32 PM

TextEdit draws the spell checker dotted underline on top too, though it's also with transparency:

Screenshot 2025-12-14 at 6 15 12 PM

pkryger:
> > an alternative could be to add gaps in the underlines.
> 
> Yeah, I do think that looks pretty good. TextEdit does the same:
> 
> Screenshot 2025-12-14 at 6 08 06 PM

I agree - this one looks smart to me. 

> but it involves typography beyond my knowledge since it requires looking at the metrics of each character. This is probably something that AppKit or CoreText handles "automatically" but I don't think that can be used for Emacs because it would mean giving up layout control.

Yeah - giving up the layout control would be undesirable. I can see that iTem2 does manipulate underline weight to achieve effect of a gap, or at least this is what I understand from [the code](https://gitlab.com/gnachman/iterm2/-/blob/44aad6ecf093a50033baefe97ab8a650daba172e/sources/Metal/Shaders/iTermTextShaderCommon.metal#L121-234). Perhaps similar trick could be employed in Emacs?

jdtsmith:
@woolsweater any luck with swapping the order so that underlines appear beneath the glyphs?  I guess that may not be as easy as it sounds (not much is with emacs display code).

woolsweater:
> @woolsweater any luck with swapping the order so that underlines appear beneath the glyphs? I guess that may not be as easy as it sounds (not much is with emacs display code).

No, I have not gotten to trying it yet. (Famous last words but) I don't think it will be too complex to change the order. That said, I am no longer sure if it is actually the right choice, given that Pages and TextEdit (for spelling) draw them on top.

jdtsmith:
OK, maybe you are right, could be a can of worms.  For drawing on top, I kind of like the alpha drawing, perhaps configured via a `mac-underline-alpha` user var (à la `mac-underwave-thickness`).  I wonder if it makes sense to deprecate the latter in favor of a more general `mac-underline-thickness` that applies to _all_ the styles?

woolsweater:
> I understand. It's possible this point is moot, since the majority of unique code here (many 10s of thousands of lines) is from Yamamoto-san , who has not assigned copyright for it. If it looks like that situation will not change, we may drop the requirement.

@jdtsmith Have you made a final decision on this point? After a year, I think we should resolve this PR one way or the other.

I would love to be able to contribute more to this project (especially with 31.1 now in the wild) so I hope we can proceed without the FSF copyright assignment.

jdtsmith:
Sorry for the long delay.  I'm now quite happy for this to go in, but at this point we should re-target `master`, which I hope @simmsb can find time to update to the v31 release, then we will cut an `emacs-mac-31` branch.

Also, I have a fairly major redesign of `emacs-mac` in the works.  The structure will I hope be a bit easier to grok and hence contribute to.  I'll be very thankful for contributions there (no FSF assignment needed).  Stay tuned...

woolsweater:
Great news, thanks! I will get this branch updated as soon as I can.

woolsweater:
@jdtsmith I don't see a "master" branch (or any similar name) in this Github repo; what would you like me to rebase this onto? Am I waiting for the merge you mentioned to happen first?

jdtsmith:
I think the `emacs-mac-31` branch is the right one.  It has been merged from upstream so it is ready.

PR #91, opened 2025-09-10
feat: Add transparent title bar

### LINKED ISSUE #83: Add support for transparent titlebar
It would be awesome if this build could support transparent titlebars, like the `ns-transparent-titlebar` variable in [homebrew-emacs-plus](https://github.com/d12frosted/homebrew-emacs-plus).

pkryger:
The implementation mechanic is inspired by the one in NS (for
`ns-transparent-titlebar`).  However, a GUI thread is used to actually
apply the value to a frame (window in macOS parlance).  This has a
slight difference, where the value can be set in a model of a frame, but
title bar transparency will not be rendered in the GUI.

I have tested it with `emacs -Q`, then:

``` emacs-lisp
(make-frame '((mac-transparent-titlebar . t)
              (background-color . "green")))
;; => a new frame with a green background and green titlebar has been created; close the frame

(let ((default-frame-alist '((mac-transparent-titlebar . t))))
  (make-frame '((background-color . "blue"))))
;; => a new frame with a blue background and blue titlebar has been created; close the frame

(set-frame-parameter nil 'background-color "yellow")
;; => background has been changed to yellow and titlebar remained in original color
(set-frame-parameter nil 'mac-transparent-titlebar t)
;; => titlebar color has changed to yellow
(set-frame-parameter nil 'mac-transparent-titlebar nil)
;; => titlebar color has changed to original one
```

fixes: #83

### DIFF
--- a/doc/emacs/macport.texi
+++ b/doc/emacs/macport.texi
@@ -539,6 +539,13 @@ title/tool/tab bars look slightly colored with the value of the
 @code{scroll-bar-background} frame parameter (or the frame background
 color) if the frame is focused and not in fullscreen.
 
+@vindex mac-transparent-titlebar
+ The title bar transparency can be controlled by the value of the
+@code{mac-transparent-titlebar} frame parameter.  If it is @code{nil},
+which is the default then title bar uses system color for its
+background.  If it is non-@code{nil} then the frame background color is
+used instead.
+
 @node Mac Fullscreen
 @section Fullscreen support on the Mac Port
 @cindex fullscreen support (Mac port)
--- a/doc/lispref/frames.texi
+++ b/doc/lispref/frames.texi
@@ -2307,6 +2307,7 @@ application's window.  (It is not certain this will be implemented; try
 it and see if it works.)
 @end ignore
 
+@ignore
 @vindex ns-appearance@r{, a frame parameter}
 @item ns-appearance
 Only available on macOS, if set to @code{dark} draw this frame's
@@ -2318,6 +2319,9 @@ background.
 
 @vindex ns-transparent-titlebar@r{, a frame parameter}
 @item ns-transparent-titlebar
+@end ignore
+@vindex mac-transparent-titlebar@r{, a frame parameter}
+@item mac-transparent-titlebar
 Only available on macOS, if non-@code{nil}, set the titlebar and
 toolbar to be transparent.  This effectively sets the background color
 of both to match the Emacs background color.
--- a/doc/lispref/frames.texi
+++ b/doc/lispref/frames.texi
@@ -2307,7 +2307,6 @@ application's window.  (It is not certain this will be implemented; try
 it and see if it works.)
 @end ignore
 
-@ignore
 @vindex ns-appearance@r{, a frame parameter}
 @item ns-appearance
 Only available on macOS, if set to @code{dark} draw this frame's
@@ -2319,9 +2318,6 @@ background.
 
 @vindex ns-transparent-titlebar@r{, a frame parameter}
 @item ns-transparent-titlebar
-@end ignore
-@vindex mac-transparent-titlebar@r{, a frame parameter}
-@item mac-transparent-titlebar
 Only available on macOS, if non-@code{nil}, set the titlebar and
 toolbar to be transparent.  This effectively sets the background color
 of both to match the Emacs background color.
--- a/src/frame.c
+++ b/src/frame.c
@@ -997,6 +997,8 @@ make_frame (bool mini_p)
 #ifdef NS_IMPL_COCOA
   f->ns_appearance = ns_appearance_system_default;
   f->ns_transparent_titlebar = false;
+#else
+  f->mac_transparent_titlebar = true;
 #endif
 #endif
   f->select_mini_window_flag = false;
@@ -4073,6 +4075,8 @@ static const struct frame_parm_table frame_parms[] =
 #ifdef NS_IMPL_COCOA
   {"ns-appearance",		SYMBOL_INDEX (Qns_appearance)},
   {"ns-transparent-titlebar",	SYMBOL_INDEX (Qns_transparent_titlebar)},
+#else
+  {"mac-transparent-titlebar",  SYMBOL_INDEX (Qmac_transparent_titlebar)},
 #endif
 };
 
@@ -6427,6 +6431,8 @@ syms_of_frame (void)
 #ifdef NS_IMPL_COCOA
   DEFSYM (Qns_appearance, "ns-appearance");
   DEFSYM (Qns_transparent_titlebar, "ns-transparent-titlebar");
+#else
+  DEFSYM (Qmac_transparent_titlebar, "mac-transparent-titlebar");
 #endif
 
   DEFSYM (Qalpha, "alpha");
--- a/src/frame.c
+++ b/src/frame.c
@@ -999,7 +999,7 @@ make_frame (bool mini_p)
   f->ns_transparent_titlebar = false;
 #endif
 #ifdef HAVE_MACGUI
-  f->mac_transparent_titlebar = true;
+  f->mac_transparent_titlebar = false;
 #endif
 #endif
   f->select_mini_window_flag = false;
--- a/src/frame.c
+++ b/src/frame.c
@@ -997,7 +997,8 @@ make_frame (bool mini_p)
 #ifdef NS_IMPL_COCOA
   f->ns_appearance = ns_appearance_system_default;
   f->ns_transparent_titlebar = false;
-#else
+#endif
+#ifdef HAVE_MACGUI
   f->mac_transparent_titlebar = true;
 #endif
 #endif
@@ -4075,7 +4076,8 @@ static const struct frame_parm_table frame_parms[] =
 #ifdef NS_IMPL_COCOA
   {"ns-appearance",		SYMBOL_INDEX (Qns_appearance)},
   {"ns-transparent-titlebar",	SYMBOL_INDEX (Qns_transparent_titlebar)},
-#else
+#endif
+#ifdef HAVE_MACGUI
   {"mac-transparent-titlebar",  SYMBOL_INDEX (Qmac_transparent_titlebar)},
 #endif
 };
@@ -6431,7 +6433,8 @@ syms_of_frame (void)
 #ifdef NS_IMPL_COCOA
   DEFSYM (Qns_appearance, "ns-appearance");
   DEFSYM (Qns_transparent_titlebar, "ns-transparent-titlebar");
-#else
+#endif
+#ifdef HAVE_MACGUI
   DEFSYM (Qmac_transparent_titlebar, "mac-transparent-titlebar");
 #endif
 
--- a/src/frame.h
+++ b/src/frame.h
@@ -744,7 +744,8 @@ struct frame
   /* NSAppearance theme used on this frame.  */
   enum ns_appearance_type ns_appearance;
   bool_bf ns_transparent_titlebar;
-#else
+#endif
+#ifdef HAVE_MACGUI
   bool_bf mac_transparent_titlebar;
 #endif
 
--- a/src/frame.h
+++ b/src/frame.h
@@ -744,6 +744,8 @@ struct frame
   /* NSAppearance theme used on this frame.  */
   enum ns_appearance_type ns_appearance;
   bool_bf ns_transparent_titlebar;
+#else
+  bool_bf mac_transparent_titlebar;
 #endif
 
 #ifdef HAVE_TEXT_CONVERSION
--- a/src/macappkit.m
+++ b/src/macappkit.m
@@ -2566,6 +2566,10 @@ static void mac_move_frame_window_structure_1 (struct frame *, int, int);
 
   window.contentView = visualEffectView;
   MRC_RELEASE (visualEffectView);
+
+  if ([window respondsToSelector:@selector(titlebarAppearsTransparent)])
+    [window setTitlebarAppearsTransparent:FRAME_MAC_TRANSPARENT_TITLEBAR(f)];
+
   FRAME_BACKGROUND_ALPHA_ENABLED_P (f) = true;
   if (FRAME_MAC_DOUBLE_BUFFERED_P (f))
     {
@@ -4284,6 +4288,16 @@ mac_set_frame_window_modified (struct frame *f, bool modified)
   NSWindow *window = FRAME_MAC_WINDOW_OBJECT (f);
 
   mac_within_gui (^{[window setDocumentEdited:modified];});
+}
+
+void
+mac_set_frame_window_transparent_titlebar (struct frame *f, bool transparent)
+{
+  NSWindow *window = FRAME_MAC_WINDOW_OBJECT (f);
+
+  mac_within_gui (^{
+      if ([window respondsToSelector: @selector(titlebarAppearsTransparent)])
+	[window setTitlebarAppearsTransparent:transparent];});
 }
 
 void
--- a/src/macfns.c
+++ b/src/macfns.c
@@ -5273,7 +5273,8 @@ frame_parm_handler mac_frame_parm_handlers[] =
 #ifdef NS_IMPL_COCOA
   NULL, /* ns_set_apperance */
   NULL, /* ns_tranaprent_titlebar */
-#else
+#endif
+#ifdef HAVE_MACGUI
   mac_set_transparent_titlebar,
 #endif
 };
--- a/src/macfns.c
+++ b/src/macfns.c
@@ -1215,6 +1215,20 @@ mac_set_override_redirect (struct frame *f, Lisp_Object new_value, Lisp_Object o
     }
 }
 
+static void
+mac_set_transparent_titlebar (struct frame *f, Lisp_Object new_value, Lisp_Object old_value)
+{
+  if (!EQ (new_value, old_value))
+    {
+      bool transparent = NILP (new_value) ? false : true;
+      FRAME_MAC_TRANSPARENT_TITLEBAR (f) = transparent;
+
+      block_input ();
+      mac_set_frame_window_transparent_titlebar (f, transparent);
+      unblock_input ();
+    }
+}
+
 /* Functions called only from `gui_set_frame_parameters'
    to set individual parameters.
 
@@ -2529,6 +2543,12 @@ DEFUN ("x-create-frame", Fx_create_frame, Sx_create_frame,
   mac_default_scroll_bar_color_parameter (f, parms, Qscroll_bar_background,
 					  "scrollBarBackground",
 					  "ScrollBarBackground", false);
+
+  tem = gui_display_get_arg (dpyinfo, parms, Qmac_transparent_titlebar,
+			     NULL, NULL, RES_TYPE_BOOLEAN);
+  FRAME_MAC_TRANSPARENT_TITLEBAR (f) = !NILP (tem) && !EQ (tem, Qunbound);
+  store_frame_param (f, Qmac_transparent_titlebar,
+		     FRAME_MAC_TRANSPARENT_TITLEBAR (f) ? Qt : Qnil);
 
   /* Init faces before gui_default_parameter is called for the
      scroll-bar-width parameter because otherwise we end up in
@@ -5247,7 +5267,15 @@ frame_parm_handler mac_frame_parm_handlers[] =
   gui_set_no_special_glyphs,
   gui_set_alpha_background,
   NULL, /* mac_set_use_frame_synchronization */
+#ifdef HAVE_X_WINDOWS
   NULL, /* mac_set_shaded */
+#endif
+#ifdef NS_IMPL_COCOA
+  NULL, /* ns_set_apperance */
+  NULL, /* ns_tranaprent_titlebar */
+#else
+  mac_set_transparent_titlebar,
+#endif
 };
 
 void
--- a/src/macfns.c
+++ b/src/macfns.c
@@ -5267,16 +5267,13 @@ frame_parm_handler mac_frame_parm_handlers[] =
   gui_set_no_special_glyphs,
   gui_set_alpha_background,
   NULL, /* mac_set_use_frame_synchronization */
-#ifdef HAVE_X_WINDOWS
-  NULL, /* mac_set_shaded */
-#endif
-#ifdef NS_IMPL_COCOA
-  NULL, /* ns_set_apperance */
-  NULL, /* ns_tranaprent_titlebar */
-#endif
-#ifdef HAVE_MACGUI
+  /* Skipping the following parameters from frame_parms that are not
+     compiled when HAVE_MACGUI is defined:
+     shaded
+     ns-appearance
+     ns-transparent-titlebar
+   */
   mac_set_transparent_titlebar,
-#endif
 };
 
 void
--- a/src/macterm.h
+++ b/src/macterm.h
@@ -339,6 +339,9 @@ struct mac_output
 #define FRAME_MAC_DOUBLE_BUFFERED_P(f) \
   ((f)->output_data.mac->double_buffered_p)
 
+#define FRAME_MAC_TRANSPARENT_TITLEBAR(f) \
+  ((f)->mac_transparent_titlebar)
+
 /* This gives the mac_display_info structure for the display F is on.  */
 #define FRAME_DISPLAY_INFO(f) ((void) (f), (&one_mac_display_info))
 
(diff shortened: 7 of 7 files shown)


jdtsmith:
Open to including something if we can work out the details.  What do you mean by:

>  the value can be set in a model of a frame, but title bar transparency will not be rendered in the GUI.

doc/lispref/frames.texi
```diff
@@ -2318,6 +2319,9 @@ background.
 
 @vindex ns-transparent-titlebar@r{, a frame parameter}
 @item ns-transparent-titlebar
+@end ignore
```
jdtsmith: We need to keep all the doc inside `doc/emacs/macport.texi`.

src/frame.c
```diff
@@ -997,6 +997,8 @@ make_frame (bool mini_p)
 #ifdef NS_IMPL_COCOA
   f->ns_appearance = ns_appearance_system_default;
   f->ns_transparent_titlebar = false;
+#else
+  f->mac_transparent_titlebar = true;
```
jdtsmith: Here and elsewhere, we need to gate any changes behind `#ifdef HAVE_MACGUI`.  Ideally we keep all patches to non-mac-exclusive files as limited as possible.

src/frame.c
```diff
@@ -4073,6 +4075,8 @@ static const struct frame_parm_table frame_parms[] =
 #ifdef NS_IMPL_COCOA
   {"ns-appearance",		SYMBOL_INDEX (Qns_appearance)},
   {"ns-transparent-titlebar",	SYMBOL_INDEX (Qns_transparent_titlebar)},
+#else
+  {"mac-transparent-titlebar",  SYMBOL_INDEX (Qmac_transparent_titlebar)},
```
jdtsmith, line 4081: Same here.

src/frame.c
```diff
@@ -6427,6 +6431,8 @@ syms_of_frame (void)
 #ifdef NS_IMPL_COCOA
   DEFSYM (Qns_appearance, "ns-appearance");
   DEFSYM (Qns_transparent_titlebar, "ns-transparent-titlebar");
+#else
+  DEFSYM (Qmac_transparent_titlebar, "mac-transparent-titlebar");
```
jdtsmith, line 6438: and here.

src/frame.h
```diff
@@ -744,6 +744,8 @@ struct frame
   /* NSAppearance theme used on this frame.  */
   enum ns_appearance_type ns_appearance;
   bool_bf ns_transparent_titlebar;
+#else
+  bool_bf mac_transparent_titlebar;
```
jdtsmith, line 749: ditto

pkryger:
> Open to including something if we can work out the details.  

Let me know what else you think should be done. 

> What do you mean by:
> 
> >  the value can be set in a model of a frame, but title bar transparency will not be rendered in the GUI.

I think I meant to describe a situation when `(set-frame-parameter nil 'mac-transparent-titlebar t)` is called and there is no support for rendering transparent titlebar in the underlying window (macOS). In such a case subsequent call to `frame-parameter' would yield `t' but the titlebar will remain opaque.

jdtsmith:
> and there is no support for rendering transparent titlebar in the underlying window (macOS).

Is that some particular version of MacOS which does not support transparent titlebar?  I'm totally ignorant about this feature.

src/macappkit.m
```diff
@@ -2566,6 +2566,10 @@ - (void)setupWindow
 
   window.contentView = visualEffectView;
   MRC_RELEASE (visualEffectView);
+
+  if ([window respondsToSelector:@selector(titlebarAppearsTransparent)])
+    [window setTitlebarAppearsTransparent:FRAME_MAC_TRANSPARENT_TITLEBAR(f)];
+
```
jdtsmith, line 2572: Does this method get run whenever the frame parameter is changed, or only when a new frame is created?

src/macfns.c
```diff
@@ -5247,7 +5267,15 @@ frame_parm_handler mac_frame_parm_handlers[] =
...
   NULL, /* mac_set_use_frame_synchronization */
+#ifdef HAVE_X_WINDOWS
   NULL, /* mac_set_shaded */
+#endif
+#ifdef NS_IMPL_COCOA
+  NULL, /* ns_set_apperance */
+  NULL, /* ns_tranaprent_titlebar */
+#else
+  mac_set_transparent_titlebar,
+#endif
```
jdtsmith: This is one of our "own" files, so we do not need to worry about NS here.

src/macfns.c
```diff
@@ -1215,6 +1215,20 @@ mac_set_override_redirect (struct frame *f, Lisp_Object new_value, Lisp_Object o
...
+    {
+      bool transparent = NILP (new_value) ? false : true;
+      FRAME_MAC_TRANSPARENT_TITLEBAR (f) = transparent;
+
+      block_input ();
+      mac_set_frame_window_transparent_titlebar (f, transparent);
+      unblock_input ();
+    }
+}
+
```
jdtsmith, line 1231: Why do we need two versions of roughly the same function, one as a method, one here?

src/frame.c
jdtsmith: This is because we don't have the luxury of being a core port, so changes made upstream can conflict with any changes we make in this port to shared files (like `frame.c`)

pkryger:
> > and there is no support for rendering transparent titlebar in the underlying window (macOS).
> 
> Is that some particular version of MacOS which does not support transparent titlebar? I'm totally ignorant about this feature.

I don't know that either. The NS implementation has an extra preprocessor guard for the transparent titlebar stuff: `MAC_OS_X_VERSION_MAX_ALLOWED >= 101000`. I assumed that this true for all platforms nowadays - my 11 years old MacBook has its support ended on BigSur 11.7.

There's also the `respondsToSelector` dynamic check, but given [the documentation](https://developer.apple.com/documentation/appkit/nswindow/titlebarappearstransparent) I guess it will be true on all platforms with macOS 10.10 and above. Note: I don't know almost anything about macOS/NS/Objective-C, but I guessed this is how `respondsToSelector` works, i.e., when instance variable is compiled it yields true.

jdtsmith:
Have you signed papers with FSF and can assign copyright if needed?

pkryger:
> Have you signed papers with FSF and can assign copyright if needed?

I did signed papers a while back. Happy to assign when requested.

jdtsmith (commented):
A few comments.

src/macfns.c
pkryger: I didn't had the NS bit in my first implementation, but I found it difficult to count member variables and their handler functions, especially when there are multiple include guards. I did it on purpose to have exactly the same flags to guard handler functions as member variables defined in `frame.c`, as per [the comment](https://github.com/jdtsmith/emacs-mac/blob/f541b48b1496deeae53cb36862b994ea08d0230d/src/macfns.c#L5195-L5196). With this approach one can open buffers side by side and just see if the variable/handler are in the same place, not really needing to know how include guards are resolved.

src/macappkit.m
pkryger: As far as I understand this is only on frame creation. The `mac_set_transparent_titlebar` is run as a handler for the `mac-transparent-tiltebar` frame parameter value change.

src/macfns.c
pkryger: I don't think I fully understood which bit of code did you had in mind. Assuming that you meant the `mac_set_frame_window_transparent_titlebar`, then answer is as follows. The latter is Objective-C code and it follows similar functions (for example `mac_set_frame_window_title`). The former `mac_set_transparent_titlebar` is a handler for a model value change in lisp.

src/macfns.c
jdtsmith: Not sure I understand the logic here, but you'll note that `NS_IMPL_COCOA` is not mentioned anywhere else in this file, and for good reason.  We don't want to give the impression that this file is used by the NS build.  Note that _some_ files (`macfont.[ch]`) are used for both NS and Mac.

If_ you think "space alignment" is a needed for comparing between builds, you can use a comment to achieve this.

doc/emacs/macport.texi
```diff
@@ -539,6 +539,13 @@ title/tool/tab bars look slightly colored with the value of the
 @code{scroll-bar-background} frame parameter (or the frame background
 color) if the frame is focused and not in fullscreen.
 
+@vindex mac-transparent-titlebar
+ The title bar transparency can be controlled by the value of the
+@code{mac-transparent-titlebar} frame parameter.  If it is @code{nil},
+which is the default then title bar uses system color for its
+background.  If it is non-@code{nil} then the frame background color is
+used instead.
+
```
jdtsmith, line 548: I think it has confused me that this is called "transparent". It doesn't allow seeing anything _behind_ the titlebar, yes?  So perhaps `mac-matched-titlebar-color` or `mac-contiguous-titlebar` or something is a more appropriate name?  Especially because various X builds do allow various forms of (real) transparency.

src/frame.c
jdtsmith: Also, does this imply we enable transparent titlebar by default?  I don't think we should do that.

pkryger: Good catch!

doc/emacs/macport.texi
pkryger: > I think it has confused me that this is called "transparent". It doesn't allow seeing anything _behind_ the titlebar, yes?

I am no macOS developer and I am not sure why the feature is called `titlebarAppearsTransparent` in [AppKit API](https://developer.apple.com/documentation/appkit/nswindow/titlebarappearstransparent?language=objc). I guess the answer depends on definition of "anything". If by that you meant non Emacs GUI widgets then yes - it doesn't allow to see anything below Emacs. However, an interpretation here may be that this strictly for titlebar, and below it there's a frame background.

> So perhaps `mac-matched-titlebar-color` or `mac-contiguous-titlebar` or something is a more appropriate name? Especially because various X builds do allow various forms of (real) transparency.

The feature has been named "natural titlebar" in [railwaycat/emacsmacport](https://github.com/railwaycat/homebrew-emacsmacport/wiki/Natural-Title-Bar), and if I understand correctly the name was taken from [emacs-plus](https://github.com/d12frosted/homebrew-emacs-plus/commit/f5b345d9b4f21ebf48990feda07f555b468cda30). However the latter [dropped this feature](https://github.com/d12frosted/homebrew-emacs-plus/commit/52192f0293b52996cde4d8c7908da6d0c9aaec34) in favour of using the `ns_transparent_titlebar`. I went for that name to keep the visual effect consistent with NS uses, as well as this was the name requested in the original issue (#83). While I appreciate the confusion it may cause (vide the definition of "anything") I am not sure if this would make sense to create a new name for the feature. I am happy to change it if you think it will be better (with dropping another name to consider `mac-titlebar-uses-frame-background`)

jdtsmith: OK, I guess there is a precedent to respect here.

src/macfns.c
jdtsmith: So our version of the `frame_parm_handlers` has fewer slots that NS's?

pkryger: It does. I added the `NS_IMPL_COCOA` guard to make this bit of code follow what's defined in frame.c automatically. But I can see the argument for bot doing so. Have just left comment/marker for future changes.

src/macfns.c
```diff
@@ -5247,7 +5267,14 @@ frame_parm_handler mac_frame_parm_handlers[] =
   gui_set_no_special_glyphs,
   gui_set_alpha_background,
   NULL, /* mac_set_use_frame_synchronization */
+#ifdef HAVE_X_WINDOWS
```
jdtsmith: Sorry for my confusion here.  Why do we need a new `HAVE_X_WINDOWS` here, which has the effect of removing `mac_set_shaded`?

NS has the following final entries in the structure:

https://github.com/jdtsmith/emacs-mac/blob/4a522f417775a539b28a6a0b3a47af7994ed39d8/src/nsfns.m#L1111-L1117

So it seems like we should just do:

```c
  gui_set_no_special_glyphs,
  gui_set_alpha_background,
  NULL, /* mac_set_use_frame_synchronization */
  NULL, /* mac_set_shaded */
  mac_set_transparent_titlebar
```

for symmetry with NS' `frame_parm_handler` structure, in case we evolve future features like `set_appearance` (though probably not as relevant nowadays).

Also, you don't need to wrap this one in `HAVE_MACGUI`, since it's a file used only by us.   Note that NS also has to support GNUSTEP, which is why it needs to check `NS_IMPL_COCOA`.

pkryger: Believe it or not, but that was my very first implementation. Unfortunately, this was prohibiting changing the titlebar transparency on existing frame, e.g., the calls to `set-frame-parameter` with the yellow background didn't work. When searching for the cause, my knee jerk reaction was to the same guards in code that is relevant to the `frame_parm` structure, i.e., to make this bit of code slightly harder to be misused by humans (computers are brilliant tracking what is defined and what's not 😉).

I have just re-checked: Taking out `HAVE_X_WINDOWS` makes the test to fail. I can take out the define and the placeholder for `mac_set_shaded` (the test pass that way), but that doesn't seem right. What's your opinion?

(FWIW: I compiled after `./configure --with-mac --enable-mac-app --with-modules --with-tree-sitter --without-native-compilation`)

jdtsmith: It would be better to understand this fully before patching.  Your patch effectively has:

```c
gui_set_no_special_glyphs,
gui_set_alpha_background,
NULL, /* mac_set_use_frame_synchronization */
mac_set_transparent_titlebar
```

and right now we have:

```c
  gui_set_no_special_glyphs,
  gui_set_alpha_background,
  NULL, /* mac_set_use_frame_synchronization */
  NULL, /* mac_set_shaded */
```

I think for your original version to work (with a longer table, like NS), you'll need to update `frame.c`, possibly to add something relevant to `frame_parm_table`.  Take a look at the comment above where that is defined. As that hints, a lisp symbol for the frame parameter gets a secret property indexing the frame param table:

```
(symbol-plist 'alpha-background) ; (x-frame-parameter 48)
```

and this is what's used to lookup what function to call.  Just follow the trail of crumbs and you should be able to figure it out!

pkryger: Actually, I didn't know the exact mechanism either.  I have just followed what
has been done for NS.  However, a bit of code investigation shows, that
the `x_frame-parameter` is dynamically added to symbols in array `frame_parms` on
Emacs startup by [function `syms_of_frame`](https://github.com/jdtsmith/emacs-mac/blob/7146872183cae537b9e024d85bdce303fa9b86d3/src/frame.c#L6489-L6499) ([called from `main`/`emacs_main`](https://github.com/jdtsmith/emacs-mac/blob/7146872183cae537b9e024d85bdce303fa9b86d3/src/emacs.c#L2388)).
This index is later used to find relevant handler function, by [indexing into a
relevant handlers table](https://github.com/jdtsmith/emacs-mac/blob/7146872183cae537b9e024d85bdce303fa9b86d3/src/frame.c#L4413-L4414) (`mac_frame_parm_handlers` in case of Emacs mac).

On a Emacs mac 30, that I build a few weeks ago (from `emacs-mac-30_1_exp`) I
can see that `shaded` is not supported:
``` emacs-lisp
(symbol-plist 'mac-transparent-titlebar)
;; nil
(symbol-plist 'shaded)
;; nil
```

When compiled with patch in this PR, I can see that
the `mac-transpared-titlebar` has an index into the table, while `shaded` is
still not supported:
``` emacs-lisp
(symbol-plist 'mac-transparent-titlebar)
;; (x-frame-parameter 50)
(symbol-plist 'shaded)
;; nil
```

I have confirmed that this is an expected result by planting the following canary code in `frame.c`:
``` c
#ifdef HAVE_X_WINDOWS
#error "canary has x windows"
#else
#error "canary does not have x window"
#endif
```

And indeed the subsequent `make` fails with the following diagnostic:
``` text
frame.c:4089:2: error: "canary does not have x window"
```

As for whether it was a bug or no I don't have a clear answer.  The last
supported frame parameter is `use-frame-synchronization` with index 49.  This
corresponds to [the first `NULL` handler in `mac_frame_parm_handlers`](https://github.com/jdtsmith/emacs-mac/blob/7146872183cae537b9e024d85bdce303fa9b86d3/src/macfns.c#L5249).  The
second one (meant for `shaded`) has been unused (at least for my build).  So the
table was a few bytes too long in the worst case (the superfluous `NULL`).

Alternatively, perhaps `HAVE_X_WINDOW` can be defined for some builds on macOS.
Maybe on a nix build? Or that used to be a case in the past?.  In such a case
the NULL was there preventing reading a handler address from beyond
the `mac_frame_parm_handler` storage.  So this extra `NULL` is just a source of
confusion.  Or a learning opportunity 🤔.  In conclusion I think I would be for
keeping the `NULL` for `shaded` under the guard in `mac_frame_parm_handler`.

pkryger: Actually, I think it's impossible to build on macOS with `HAVE_X_WINDOWS` defined.  At least this is what I am getting from [how it is set](https://github.com/jdtsmith/emacs-mac/blob/7146872183cae537b9e024d85bdce303fa9b86d3/configure.ac#L3295-L3326) in response to [`window_system` is set](https://github.com/jdtsmith/emacs-mac/blob/7146872183cae537b9e024d85bdce303fa9b86d3/configure.ac#L2862-L2866).  Unless I am missing something the question remains, do you prefer to have the guard (I still vote for that), or just remove the superfluous `shaded` handler.

jdtsmith: `HAVE_X_WINDOWS` is just not relevant for and has no place this file. There is no such thing as a `mac` build with `HAVE_X_WINDOWS`.  Including it here is the same as `#if 0`.  I.e. a no-op.  And so confusing for future readers. 

Unfortunately, I don't really know what you mean by guard.  What I think we should have is your "original solution," which I came to independently:

```
  gui_set_no_special_glyphs,
  gui_set_alpha_background,
  NULL, /* mac_set_use_frame_synchronization */
  NULL, /* mac_set_shaded */
  mac_set_transparent_titlebar
```

This means leaving those two NULLS where they are, and making sure `(symbol-plist 'mac-transparent-titlebar)` points to the correct slot in this array, and that nothing points past the _end_ of the array.

pkryger: > `HAVE_X_WINDOWS` is just not relevant for and has no place this file. There is no such thing as a `mac` build with `HAVE_X_WINDOWS`. Including it here is the same as `#if 0`. I.e. a no-op. And so confusing for future readers.

I will remove it.

> Unfortunately, I don't really know what you mean by guard.

Sorry, I meant `ifdef`.

> What I think we should have is your "original solution," which I came to independently:
> 
> ```
>   gui_set_no_special_glyphs,
>   gui_set_alpha_background,
>   NULL, /* mac_set_use_frame_synchronization */
>   NULL, /* mac_set_shaded */
>   mac_set_transparent_titlebar
> ```
> This means leaving those two NULLS where they are, and making sure `(symbol-plist 'mac-transparent-titlebar)` points to the correct slot in this array, and that nothing points past the _end_ of the array.

Unfortunately, this won't work. Please see [how the indexes are assigned](https://github.com/jdtsmith/emacs-mac/blob/7146872183cae537b9e024d85bdce303fa9b86d3/src/frame.c#L6489-L6499). The line:
```
>   NULL, /* mac_set_shaded */
```
is superfluous. In the current code (i.e., with this patch) it is not doing anything besides allocating a few bytes too many. However, it stays in a way of appending any more frame parameters (I mean: adding after `use-frame-synchroisation`), since this `NULL` occupies a slot selected by the index assigned to a first appended parameter (and all other parameters are off by one in consequence).

jdtsmith: I have looked at how the indexes are assigned, but I admit I don't see that NULL poses any special problem, and there are other NULLs e.g. for NS that don't lead to problems, right?  The index continues to increment. In any case I suppose dropping one
empty entry is not so bad.  Maybe a comment warning about slot alignment with other builds.  Thanks for digging in here.

pkryger: I don't think I get what you meant by:

> The index continues to increment. In any case I suppose dropping one empty
> entry is not so bad.

I think that a superfluous `NULL` in `mac_frame_parm_handlers` poses a problem,
when the superfluous `NULL` is not the last entry.  Please consider the
following, small, example with a superfluous `NULL` entry
in `mac_frame_parm_handlers` (e.g., for `shaded` not supported in macOS build):


``` c
static const struct frame_parm_table frame_parms[] =
{
  {"auto-raise",		SYMBOL_INDEX (Qauto_raise)},
#ifdef HAVE_X_WINDOWS
  {"shaded",			SYMBOL_INDEX (Qshaded)},
#endif
#ifdef HAVE_MACGUI
  {"mac-transparent-titlebar",  SYMBOL_INDEX (Qmac_transparent_titlebar)},
#endif
};

frame_parm_handler mac_frame_parm_handlers[] =
{
  gui_set_autoraise,
  NULL, /* the superfluous entry, e.g., for mac_shaded */
  mac_set_transparent_titlebar,
};
```

On macOS, this code is compiled code with undefined `HAVE_X_WINDOS` and
defined `HAVE_MACGUI` defined.  The resulting data structures in compiled data
will be (pseudo-code, values of `iQ...` come from generated `src/globals.h`):

``` assembly
_mac_frame_parm_handlers:
    _gui_set_autoraise             // index: 0
    0                              // index: 1
    _mac_set_transparent_titlebar  // index: 2
_frame_parm_table:
    "auto-rise".                  // index: 0
    349                           // iQauto_raise
    "mac-transparent-titlebar     // index: 1
    1078                          // iQmac_transparent_titlebar
```

At Emacs start [the index assigning code](https://github.com/jdtsmith/emacs-mac/blob/7146872183cae537b9e024d85bdce303fa9b86d3/src/frame.c#L6489-L6499) will execute something that resembles the following unrolled pseudo code:

``` c
Lisp_object v;
i = 0;
v = builtin_lisp_symbol (349 /* iQauto_raise */);
Fput (v, Qx_frame_parameter, i);; // i.e.,  (put 'auto_resize 'x-frame-parameter 0)
i++;
v = builtin_lisp_symbol (1078 /* iQmac_transparent_titlebar */);
Fput (v, Qx_frame_parameter, i);; // i.e.,  (put 'mac_transparent_titlebar 'x-frame-parameter 1)
```

Now, when `gui_set_frame_parameters_1` is executed, say in response
to `(set-frame-parameter 'mac-transparent-titlebar t)`, [it will execute](https://github.com/jdtsmith/emacs-mac/blob/7146872183cae537b9e024d85bdce303fa9b86d3/src/frame.c#L4410-L4414):

``` c
struct frame *f = ...;
Lisp_object val = Qt;
Lisp_object old_val = ...;

parm_index = Fget (1078 /* iQmac_transparent_titlebar */,
                   Qx_frame_parameter); // i.e., (get 'mac_transparent_titlebar 'x-frame-parameter) => parm_index = 1
if (/* data type check ommitted for brevity */
   _mac_frame_parm_handlers[1 /* parm_index */] &&
   (*(_mac_fraem_parm_handlers[1 /* parm_index */]))(f, Qt /* val */, old_val))
```
However the superfluous `NULL` will cause function pointers in `mac_frame_parm_handlers` to be off by one.

Given the above, I think that function pointer to handlers
in `mac_frame_parm_handlers` should be synchronised with a compiled version
of `frame_parm_table`.  The frame parameter `shaded` [is not compiled in](https://github.com/jdtsmith/emacs-mac/blob/7146872183cae537b9e024d85bdce303fa9b86d3/src/frame.c#L4070-L4072), when
building Emacs mac, hence there should be no entry for it
in `mac_frame_parm_handlers`.

Side note: The entry for `shaded` in `mac_frame_parm_handlers` [is defined in
current code](https://github.com/jdtsmith/emacs-mac/blob/7146872183cae537b9e024d85bdce303fa9b86d3/src/macfns.c#L5250), however I believe it to be not an issue.  The reason is that at
runtime there is no frame parameter defined, such that the frame
parameter's `x-frame-parameter` value indexes into the entry for `shaded`
in `mac_frame_parm_handlers`.  So there is no option for an user to
say `(set-frame-parameter SOMETHING)` and expect that some handler code to be
executed.  Side note to side note: Unless they define such a symbol with
appropriate `x-frame-parameter`, but I don't suppose this should ever be
handled.

jdtsmith: Thanks for the detailed analysis. 

> Given the above, I think that function pointer to handlers
in mac_frame_parm_handlers should be synchronised with a compiled version of frame_parm_table. 

I agree entirely.  Where we seem to be missing each other: I was advocating ensuring this with updated entries in both tables. But given your analysis I'm ok with the current state. Just add a comment about the offset from other builds?

jdtsmith:
Thanks for your patience and this contribution!

Issue #50, opened 2025-07-28
Starting src/emacs


gerd-moellmann:
In a terminal, run `src/emacs -Q`. With my patch to remove
CF_NOESCAPE, and with clang 20, Emacs starts, a window opens, but
keyboard input does not work.

NS has this code to make it work:
 ```
nsterm.m<master>:
 6346   if ([NSApp activationPolicy] == NSApplicationActivationPolicyProhibited) {
 6347     /* Set the app's activation policy to regular when we run outside
 6348        of a bundle.  This is already done for us by Info.plist when we
 6349        run inside a bundle.  */
 6350     [NSApp setActivationPolicy:NSApplicationActivationPolicyRegular];
 6351     [NSApp setApplicationIconImage:
 6352              [EmacsImage
 6353                allocInitFromFile:
 6354                  build_string("icons/hicolor/128x128/apps/emacs.png")]];
 6355   }
 ```
 
Adding the euivalent to emacs-mac:
```
 macappkit.m:
 1276   if ([NSApp activationPolicy] == NSApplicationActivationPolicyProhibited) {
 1277     /* Set the app's activation policy to regular when we run outside
 1278        of a bundle.  This is already done for us by Info.plist when we
 1279        run inside a bundle.  */
 1280     [NSApp setActivationPolicy:NSApplicationActivationPolicyRegular];
 1281 #if 0
 1282     NSString *filename = @"icons/hicolor/128x128/apps/emacs.png";
 1283     NSImage *image = [[NSImage alloc] initByReferencingFile:filename];
 1284     [NSApp setApplicationIconImage:image];
 1285 #endif
 1286   }
```

makes this also work in emacs-ac. The `if 0` is there because I don't
know how emacs-mac finds image files in some search patch like NS
does.

jdtsmith:
OIC.  Interesting.  But running `mac/Emacs.app/Contents/MacOS/Emacs -q` is not so bad...

gerd-moellmann:
JD Smith ***@***.***> writes:

> *
> jdtsmith left a comment (jdtsmith/emacs-mac#50)
>
> OIC. Interesting. But running mac/Emacs.app/Contents/MacOS/Emacs -q is
> not so bad...

I'd appreciate if it were added, for consistency, as someone who uses
different Emacssen. And what you wrote about sometimes the pdump in mac/
being wrong or what it was can't happen :-).

gerd-moellmann:
[...]

Issue #1, opened 2025-03-11
macOS 12.7.6 build succeeds, binary hangs


shipmints:
Curious to try emacs-mac on Emacs 30.  Thanks for giving this project a shot.

Build succeeded just fine.  It "hangs" when src/emacs -Q.

What platform(s) are you developing/testing this on?

Below is likely not useful...
```
$ uname -a
Darwin 21.6.0 Darwin Kernel Version 21.6.0: Mon Jun 24 00:56:10 PDT 2024; root:xnu-8020.240.18.709.2~1/RELEASE_X86_64 x86_64

$ clang --version
Apple clang version 14.0.0 (clang-1400.0.29.202)
Target: x86_64-apple-darwin21.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin

$ lldb src/emacs
(lldb) run -Q
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGSTOP
  * frame #0: 0x00007ff80890b93a libsystem_kernel.dylib`mach_msg_trap + 10
    frame #1: 0x00007ff80890bca8 libsystem_kernel.dylib`mach_msg + 56
    frame #2: 0x00007ff808a0f29d CoreFoundation`__CFRunLoopServiceMachPort + 319
    frame #3: 0x00007ff808a0d928 CoreFoundation`__CFRunLoopRun + 1276
    frame #4: 0x00007ff808a0cd6c CoreFoundation`CFRunLoopRunSpecific + 562
    frame #5: 0x00007ff80986bb5a Foundation`-[NSRunLoop(NSRunLoop) runMode:beforeDate:] + 216
    frame #6: 0x000000010027fb5f emacs`__mac_select_block_invoke_2.2844(.block_descriptor=<unavailable>) at macappkit.m:16992:5 [opt]
    frame #7: 0x0000000100251660 emacs`-[NSApplication(self=0x0000000102c05c00, _cmd=<unavailable>, block=<unavailable>) stopAfterCallingBlock:] at macappkit.m:501:3 [opt]
    frame #8: 0x00007ff809934107 Foundation`__NSFirePerformWithOrder + 357
    frame #9: 0x00007ff808a0e450 CoreFoundation`__CFRUNLOOP_IS_CALLING_OUT_TO_AN_OBSERVER_CALLBACK_FUNCTION__ + 23
[...]

PR #93, opened 2025-09-13
Allow starting Emacs app from src/emacs

jdtsmith:
* src/macappkit.m (install_application_handler): New function.
* src/macterm.h: Declare it.
* src/macterm.c (mac_initialize): Call it.

As proposed by @gerd-moellmann in #50.

With this, you can start the "raw" just-built `src/emacs` executable and
it will function as a full app, without needing to `make install`.

### DIFF
--- a/src/macappkit.m
+++ b/src/macappkit.m
@@ -1954,6 +1954,25 @@ emacs_windows_need_display_p (void)
 
 @end				// EmacsController
 
+void
+init_activation_policy (void)
+{
+  mac_within_gui (^{
+    if ([NSApp activationPolicy] == NSApplicationActivationPolicyProhibited)
+      {
+	/* Set the app's activation policy to regular when we run
+	   outside of a bundle.  This is already done for us by
+	   Info.plist when we run inside a bundle.  */
+	[NSApp setActivationPolicy:NSApplicationActivationPolicyRegular];
+
+	const char *file = "mac/Emacs.app/Contents/Resources/Emacs.icns";
+	NSString *path
+	  = [NSString stringWithFormat:@"%s/../%s", SDATA (Vdata_directory), file];
+	NSImage *image = [[NSImage alloc] initByReferencingFile:path];
+	[NSApp setApplicationIconImage:image];
+      }});
+}
+
 OSStatus
 install_application_handler (void)
 {
--- a/src/macterm.c
+++ b/src/macterm.c
@@ -5858,6 +5858,8 @@ mac_initialize (void)
 
   record_startup_key_modifiers ();
 
+  init_activation_policy ();
+
   unblock_input ();
 }
 
--- a/src/macterm.h
+++ b/src/macterm.h
@@ -616,6 +616,7 @@ extern CFStringRef mac_uti_create_with_mime_type (CFStringRef);
 extern CFStringRef mac_uti_create_with_filename_extension (CFStringRef);
 extern CFStringRef mac_uti_copy_filename_extension (CFStringRef);
 extern CFStringRef mac_uti_copy_mime_type (CFStringRef);
+extern void init_activation_policy (void);
 extern OSStatus install_application_handler (void);
 extern Lisp_Object mac_application_state (void);
 extern void mac_set_frame_window_title (struct frame *, CFStringRef);


jdtsmith:
I did notice that some keybindings are affected.  E.g. `C-h` seems to be caught by the OS when started from `src/emacs`.  But this problem goes away when starting like:

```
% mac/Emacs.app/Contents/MacOS/Emacs
```

perhaps due to something in the `Info.plist` file.  @gerd-moellmann do you also see that?

gerd-moellmann:
JD Smith ***@***.***> writes:

> *
> jdtsmith left a comment (jdtsmith/emacs-mac#93)
>
> I did notice that some keybindings are affected. E.g. C-h seems to be caught by the OS when started from
> src/emacs. But this problem goes away when starting like:
>
> % mac/Emacs.app/Contents/MacOS/Emacs
>
> perhaps due to something in the Info.plist file. @gerd-moellmann do you also see that?

I don't see that. What happens, also with NS, is that Emacs' window is
not automatically brought to the front, so one has to do that ny hand.
Could it be that?

(T thought I remembered a discussion about that, regarding NS, and that
it was decided not to change that, for whatever reason, but I can't
anything anymore. Maybe I'm hallucinating.)

jdtsmith:
Looks like `C-h` = backspace for me when started from `src/emacs`.  But operationally it seems starting `Emacs` from within the temporary staging app without installing works fine, because both of these now include all the `build-dir/lisp/*` directories on `load-path`.  

BTW, do you disable `native-comp` during testing runs like this?  Just omitting `--with-native-compilation`?

gerd-moellmann:
> On 13. Sep 2025, at 15:47, JD Smith ***@***.***> wrote:
> 
> 
> jdtsmith
>  left a comment 
> (jdtsmith/emacs-mac#93)
>  <https://github.com/jdtsmith/emacs-mac/pull/93#issuecomment-3288384004>
> Looks like C-h = backspace for me when started from src/emacs. But operationally it seems starting Emacs from within the temporary staging app without installing works fine, because both of these now include all the build-dir/lisp/* directories on load-path.
> 
> BTW, do you disable native-comp during testing runs like this? Just omitting --with-native-compilation?
> 
I do that frequently, yes, because the build time without native-comp are much faster. It is --with-native-compilation=no because building with native comp became the default at some point.

jdtsmith:
Thanks. Do you also skip byte-compiling all the `.el` files for speed?

I made a `git worktree` that's solely for debug testing (in which I never run `make install`) and configure there with debugging and without native comp.  Then `cd src; lldb ../mac/Emacs.app` works great.  Not clear if this patch is actually necessary for that, however.

gerd-moellmann:
JD Smith ***@***.***> writes:

> *
> jdtsmith left a comment (jdtsmith/emacs-mac#93)
>
> Thanks. Do you also skip byte-compiling all the .el files for speed?

No. A full build starting from git clean -xdf takes 1m here when I use a
config cache, so that doesn't really pay off.

(I'd also use a config cache because configure is so slow. It's a
difference light night and day :-)).

> I made a git worktree that's solely for debug testing (in which I
> never run make install) and configure there with debugging and without
> native comp. Then cd src; lldb ../mac/Emacs.app works great. Not clear
> if this patch is actually necessary for that, however.

It depends on what you have to debug. For example if you need to pass
command line flags. Or if you plan to profile with xctrace and such.

If neither is the case, using xattach might indeed be the easiest thing
to do. For that, M-x lldb RET RET with current directory being src/,
then xattach in lldb, and start Emacs.

jdtsmith:
> use a config cache because configure is so slow. 

How does that look?  Whenever you need to try a different config flag it's a bit painful.

gerd-moellmann:
JD Smith ***@***.***> writes:

> *
> jdtsmith left a comment (jdtsmith/emacs-mac#93)
>
>  use a config cache because configure is so slow.
>
> How does that look? Whenever you need to try a different config flag
> it's a bit painful.

I have something like this is my make-emacs script

worktree="$(basename $(pwd))"
config_cache="$TMPDIR/config.cache.$worktree"

# Delete the config cache file if --no-config is specified.
# Unfortunately, this makes configure print a warning, but
# that I couldn't find a way to suppress.
if [ "$no_cache" != "" ]; then
    rm -f $config_cache
    touch $config_cache
fi

# Create or use a cache.
config_flags+=(--cache-file $config_cache)

That is, configure is called with --nache-file FILE, where FILE is
derived from the name of the worktree being built. I have to use
make-emacs --no-cache only if I want to change settings, for example
CFLAGS or whatever. Configure will notice changed CFLAGS etc. quickly and
exit, and I have to make-emacs --no-cache then.

Something like that. Most% of the time I profit from the cache because I
frequently build from scratch, just to be sure, and because it's fast.

shipmints:
Greetings. I recently acquired a pretty-well equipped macBook Pro M1 Max on which I installed macOS 26/Tahoe. JD's recent email missive reminded me to give this a try again on this machine (now my main machine, now partially incompatible with the others I use). I built `emacs-mac` just fine. I went to run it out of src as in `src/emacs -Q` as always. Runs fine with `-nw` but still the same GUI hang as in https://github.com/jdtsmith/emacs-mac/issues/1 which confused me since I figured it was just me on macOS 12.7. BUT, I read this pull request and when I run it forcibly out of the `mac/Emacs.app` tree, it runs fine under GUI. Could this have been the issue all along that `src/emacs` never worked and that detail just never came up in my investigation? It wasn't mentioned in #1. Will this PR land soon to help avoid confusion in the future?

jdtsmith:
Interesting that may explain your issues.  Running "out of bundle" like that is a bit of a hack, so isn't (yet) supported (see the README).  Did you get the impression this was possible just from NS? For me this PR works, but leads to key event issues from terminal translation. 

I now think the right approach is to make the `./mac/Emacs.app`:

1. a clean install (now it's also a build dir) and
2. truly self-contained (`native-lisp/` often goes missing or is wrong-versioned).  

Then `make install` can just relocate the app bundle and compress Elisp files (and whatever else it does) fully inside it. 

Does that sound like a good solution?  If so, are you willing to work up something along these lines?

shipmints:
In the end, emacs is just a process so, yes, I expected it to work out of bundle. NS works (aside from the initial window not raising which I reported a while ago). I'm not clear on what a bundle provides over and above what's standalone, so it's hard to say I have an opinion on bundles. I'm away from home (still recovering from a hand injury and still typing one-handed yay fun) and when I return, I'll try again on one of my macOS 12 Intel machines and see if the bundle launches correctly.

As I hack on Emacs itself, I prefer running from src pretty much all the time when hacking so the bundle solution doesn't seem optimal unless I'm building for an installation (which is rare for me).

gerd-moellmann:
Stéphane Marks ***@***.***> writes:

> As I hack on Emacs itself, I prefer running from src pretty much all
> the time when hacking so the bundle solution doesn't seem optimal
> unless I'm building for an installation (which is rare for me).

FWIW, I don't notice anything weird regarding keys in my Emacs.

jdtsmith:
> In the end, emacs is just a process so, yes, I expected it to work out of bundle.  I'm not clear on what a bundle provides over and above what's standalone, so it's hard to say I have an opinion on bundles.

It's a Mac peculiarity for sure.  The bundle contains all sorts of relevant config information aside from the executable (that's why a bundle exists in the first place).  I doubt you'll find many/any other Cocoa GUI apps that can be started that out-of-bundle.  Even mouse drivers, etc. do:

```
/Library/Application Support/Logitech.localized/LogiOptionsPlus/logioptionsplus_agent.app/Contents/MacOS/logioptionsplus_agent
```

Our bundle isn't too complicated: just all the lisp files, an `Info.plist`, the app icons, code signatures, pre-compiled native-lisp (see `Frameworks/native-lisp/30_2_*`), and some localization files.  The lisp files can be located and found elsewhere of course, if you do a "non-self-contained" build.

> As I hack on Emacs itself, I prefer running from src pretty much all the time when hacking so the bundle solution doesn't seem optimal unless I'm building for an installation (which is rare for me).

Yeah I feel that use case too.  But if a normal `make` populates a fully functional `mac/Emacs.app` without doing the extra work done by `make install`, there would be no practical difference, yes.  I.e. you can just run:

```
./mac/Emacs.app/Contents/MacOS/Emacs
```

or 

```
open ./mac/Emacs.app
```

instead of

```
src/emacs
```

with no difference in build time.  That's my proposal. The key missing piece is getting a better understanding of what `make install` currently does, and making that operate exclusively on the "already built" app bundle.

shipmints:
Showing macOS ignorance... A good question is why do `NS` builds work out of `src` while `emacs-mac` builds don't and can we eliminate the need for the bundle rather than shoehorn into it?

gerd-moellmann:
Stéphane Marks ***@***.***> writes:

> *
> shipmints left a comment (jdtsmith/emacs-mac#93)
>
> Showing macOS ignorance... A good question is why do NS builds work
> out of src while emacs-mac builds don't and can we eliminate the need
> for the bundle rather than shoehorn into it?

Simple - NS has the equivalent of my patch, emacs-mac has not :-).

jdtsmith:
It's more "hack to work outside it" than "shoehorn within it".  But what's the difference in actual use?

gerd-moellmann:
JD Smith ***@***.***> writes:

> jdtsmith left a comment (jdtsmith/emacs-mac#93)
>
> It's more "hack to work outside it" than "shoehorn within it". But
> what's the difference in actual use?

For me, it's making my tooling work the same across platforms.

shipmints:
Agree. Uniformity across platforms is key to lower cognitive load and uniform tooling. Especially when contributions need to be tested for Unix, NS, Carbon (setting aside Windows which I haven't had a lot of luck with the tooling there is still pretty bad IMO).

jdtsmith:
Do either of you know: how does `src/emacs` know where to look for lisp files and native-compiled lisp files?

shipmints:
I think `load-path` defaults relative to `invocation-directory`. `native-comp-eln-load-path` is always relative to `user-emacs-directory`.

jdtsmith:
I mean the pre-compiled native-comp files:

```
% \ls -1 Frameworks/native-lisp/30_2_50-b276b136
byte-opt-9c5f25f5-5065798b.eln
bytecomp-12882072-ce55a8c4.eln
charscript-600dca1a-b361c995.eln
comp-7672a6ed-99c933df.eln
comp-common-6e17f702-570d32c2.eln
comp-cstr-ef162ef7-4beebc10.eln
comp-run-a15747ee-6eeac0b4.eln
emoji-zwj-4f682c68-423d0a74.eln
ja-dic-283bfd77-d6c74399.eln
loaddefs-gen-e8a3ad9c-b89bffc7.eln
preloaded
radix-tree-669a468d-4c01b43c.eln
titdic-cnv-765ac3be-0bd45946.eln
```

**Update**: I see that actually is on `native-comp-eln-load-path`:

```
native-comp-eln-load-path is a variable defined in ‘C source code’.

Its value is
("/.../.emacs.d/eln-cache/"
 "/Applications/Emacs.app/Contents/Frameworks/native-lisp/")
```

gerd-moellmann:
JD Smith ***@***.***> writes:

> *
> jdtsmith left a comment (jdtsmith/emacs-mac#93)
>
> Do either of you know: how does src/emacs know where to look for lisp files and native-compiled lisp files?

Does C-h v load-path and then following the link to the Elisp manual
help? The only thing I still remember is that it was kind of
complicated.

jdtsmith:
Confusing stanza in `Makefile.in`:

```
ifeq (${mac_self_contained},no)
ifeq (${ns_self_contained},no)
BIN_DESTDIR='$(DESTDIR)${bindir}/'
ELN_DESTDIR = $(DESTDIR)${libdir}/emacs/${version}/
else
BIN_DESTDIR='${ns_appbindir}/'
ELN_DESTDIR = ${ns_applibdir}/
endif
else
BIN_DESTDIR='${mac_appbindir}/'
ELN_DESTDIR = ${mac_applibdir}/
endif
```

This is the kind of stuff I wish we could eradicate.

shipmints:
I propose we merge this PR, especially as this works in NS, and we can see if the makefile needs refinement for any common case. If the bundle build is near instantaneous, I'll just leave it but ignore it.

jdtsmith:
Can you please test it first?  For me when I build with this PR and run `src/emacs` from `Terminal.app`, `C-h` is treated as `<backspace>`, `C-m` is treated as `<return>`, etc.  If we merge this as-is, this issue is sure to be opened soon.  So please let's fix it now. 

One clue is that this does _not_ happen when run from `Terminal.app` using `mac/Emacs.app/Contents/MacOS/Emacs`.  Presumably something in `Info.plist` configures this when run inside a bundle.   

Overall, though, I'm not totally convinced this is the right thing to do.  The "tiny fake plist" approach that NS uses is IMO guaranteed to leave inconsistencies such as the one I encountered.  Since bundles are also _signed_, running out-of-bundle may (sometime) lead to security warnings, etc.  All stuff we don't want to deal with, just to avoid a tiny separate test in our tooling that switches from `src/emacs` to `mac/Emacs.app/Contents/MacOS/Emacs`.  I don't love app bundles either, but the fact of the matter is that they are universal and assumed on MacOS.

shipmints:
I'll try to test it today. Can you point me to the mini plist thingy in NS? When I look at the plist files in both master and emacs-mac trees, I see roughly the same number of plist files.

jdtsmith:
It's basically doing in code (a tiny part of) what an `Info.plist` file does directly; check the code in this PR, which just mirror's NS's.  It just overrides the "app flavor" when run out of bundle, a sets the Icon.  Maybe there are a few more "fake PLIST" things we can do to help it behave a bit better.

shipmints:
Just to be super clear, when you see the keyboard translation issues launched from Terminal.app, are you launching -nw or GUI?

shipmints:
On macOS 26/xcode 26 (i.e., clang 17), I took the latest emacs-mac-30_1_exp and cherry-picked the commit from start-src-emacs and it seems to run fine for me under terminal with -Q -nw, -Q (GUI), and also with my production configuration. If I build straight from start-src-emacs, -nw works but GUI started once and then on subsequent starts hangs (killable with SIGTERM 15, required twice, so not a deadlock). 

I did not see any issues with keyboard translations but I'm not precisely sure what to look for. I typed around, pressed C-h for help, backspace for deleting previous character at point, etc. I do have an external PC-style keyboard where I am right now that I could try if you think that's of any value.

pkryger:
It is a long shot, but how about changing `Makefile` to not produce (or remove) `./src/emacs` and create a symlink into bundle instead?

jdtsmith:
GUI.  Just plain `src/emacs`.

jdtsmith:
@pkryger I did think of that as a simple solution, but no joy:

```
% src/emacs
Loading loadup.el (source)...
Dump mode: nil
Using load-path (Contents/Resources/site-lisp /Users/_/code/emacs/emacs-mac/lisp /Users/_/code/emacs/emacs-mac/lisp/emacs-lisp /Users/_/code/emacs/emacs-mac/lisp/progmodes /Users/_/code/emacs/emacs-mac/lisp/language /Users/_/code/emacs/emacs-mac/lisp/international /Users/_/code/emacs/emacs-mac/lisp/textmodes /Users/_/code/emacs/emacs-mac/lisp/vc)
Loading emacs-lisp/debug-early...
Symbol's function definition is void: file-name-sans-extension
```

shipmints:
@jdtsmith do you happen to have any macOS keyboard filters installed like karabiner that might need some accommodation?

jdtsmith:
Yep, that was it.  I wonder why it only impacts keys when GUI is run from outside the bundle.

shipmints:
Going to guess you have an exception for Emacs to not filter anything (or reduce filtering) and when Emacs is in a bundle, the filter can take effect, but not for a standalone process. There could be a karabiner setting or this could be a macOS limitation. Perhaps send them a note and see.

jdtsmith:
~~It was actually the other way around.~~  I have a Karabiner rule that adds back Emacs bindings in various locations to things like the Messages app.  It does nothing with actual GUI Emacs, but when run as `src/emacs`, the rule ~~takes effect~~ is disabled.  It included `C-h` → `BS` mapping and others.  I've disabled that for emacs and now `src/emacs` works fine (though it does not come to the front, as Gerd mentioned).  I think my issue is sufficiently rare that since it tests OK for you and Gerd both, I'm happy to merge it.  

Thanks for your patience.

jdtsmith:
Disregard that.  I'm using a Karabiner Emacs key set that is _disabled_ based on bundle id, including for:

```
"bundle_identifiers": [
                        "^org\\.gnu\\.Emacs$",
```

When run outside the bundle, `src/emacs` has no bundle identifier, so tools (like Karabiner) which match based on that id fail.  Another reason "not to do that", IMO.

pkryger:
Perhaps [trick that d12frosted/emacs-plus tap does](https://github.com/d12frosted/homebrew-emacs-plus/blob/1a5476d127145c712b0216fc68a87195b11952bc/Formula/emacs-plus%4030.rb#L215-L220) could be employed here?

jdtsmith:
Honestly not a bad idea.  Would that work for you, @gerd-moellmann and @shipmints?

gerd-moellmann:
JD Smith ***@***.***> writes:

> *
> jdtsmith left a comment (jdtsmith/emacs-mac#93)
>
> Honestly not a bad idea. Would that work for you, @gerd-moellmann and @shipmints?

IIUC, this is a bash script named "emacs" somewhere? Maybe that's
something useful for users who miss being able to start Emacs from the
command line by typing "emacs".

It would not fix things for developers who are used to src/emacs being
runnable everywhere else, though. Think lldb, xctrace, shell scripts
running benchmarks, to name some use-cases of mine.

shipmints:
I agree with @gerd-moellmann src/emacs is a binary that needs to be able to run straight away and/or under a debugger with full functionality.

jdtsmith:
> IIUC, this is a bash script named "emacs" somewhere?

Somewhere being `src/emacs`, i.e. overwriting the binary with a bash script.

Yeah, figured that might be your opinion.  One real problem with the developer habit of debugging from `src/emacs`: since `Info.plist` sets various preferences, these could differ when run out of bundle, making bug reports less valuable.  Hopefully only dev-types with too much muscle memory to adapt will do that :).

Some niceties like an `emacs` bash script on `$PATH` we should leave to packagers like brew.

BTW, you can specify the app bundle for `lldb` directly (e.g. `lldb ../mac/Emacs.app`) and it will DTRT.  That's how I debug, or more commonly just attach to the already running process post-facto (e.g. `process attach -p 29511`), which is very nice since I don't have to remember to run under lldb.

gerd-moellmann:
JD Smith ***@***.***> writes:

> *
> jdtsmith left a comment (jdtsmith/emacs-mac#93)
>
>  IIUC, this is a bash script named "emacs" somewhere?
>
> Somewhere being src/emacs, i.e. overwriting the binary with a bash
> script.

Catastrophe!

> Yeah, figured that might be your opinion. One real problem with the
> developer habit of debugging from src/emacs: since Info.plist sets
> various preferences, these could differ when run out of bundle, making
> bug reports less valuable. Hopefully only dev-types with too much
> muscle memory to adapt will do that :).
>
> Some niceties like an emacs bash script on $PATH we should leave to
> packagers like brew.
>
> BTW, you can specify the app bundle for lldb directly (e.g. lldb
> ../mac/Emacs.app) and it will DTRT. That's how I debug, or more
> commonly just attach to the already running process post-facto (e.g.
> process attach -p 29511), which is very nice since I don't have to
> remember to run under lldb.

I know, but it's different from what I have to do anywhere else, most
importantly ns and tty. I'm not concerned about the contents of
Info.plist. I have yet to see someone encountering a difference, with
the exception of this Karabiner thing.

PR #95, opened 2025-09-16
Disable NSAutoFillHeuristicController

simmsb:
MacOS 26 is bugged in that applications become progressively slower at handling events over time, this is due to some behavior by NSAutoFillHeuristicController.

See https://github.com/ghostty-org/ghostty/pull/8625 for further information.

### DIFF
--- a/src/macappkit.m
+++ b/src/macappkit.m
@@ -6451,6 +6451,8 @@ static BOOL emacsViewUpdateLayerDisabled;
     {
       NSUserDefaults *defaults = NSUserDefaults.standardUserDefaults;
 
+      if ([defaults objectForKey:@"NSAutoFillHeuristicControllerEnabled"] == nil)
+	[defaults registerDefaults:@{@"NSAutoFillHeuristicControllerEnabled" : @false}];
       if ([defaults objectForKey:@"ApplePressAndHoldEnabled"] == nil)
 	[defaults registerDefaults:@{@"ApplePressAndHoldEnabled" : @"NO"}];
     }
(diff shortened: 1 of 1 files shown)


gerd-moellmann:
Should that be `@"NO"` instead of `@"false"`?

Anyway, this also works with `@false` on Tahoe, and it gets rid of an extra process that is started without and can be seen in Activity Monitor. (Built with system Clang because LLVM from Homebrew currently does not work with Xcode 16 because of the missing SDK for macOS 26. (Which can be fixed by editing `/opt/homebrew/etc/clang/arm64-apple-darwin25.cfg`)).

jdtsmith:
Weird.  Have you been testing this out on your master branch @simmsb?  Did you ever see the mentioned "event degradation"?

simmsb:
Hey, yeah I was experiencing this issue quite often since I started using the beta of macos26 a month ago.

I'll also change this to `@"NO"`, but `@false` does seem to work.

jdtsmith:
Want to try `@"NO"`?

jdtsmith:
Can you tell if this is still needed with the fixes from upstream (see #96)?  I just pushed those up to `30.1_exp`

jdtsmith:
Ping @simmsb.

simmsb:
Hmm, it seems like one of the upstream changes, or perhaps a macos update fixed this problem.
I'm closing this for now, if someone reports stuttering then we can consider applying this.

jdtsmith:
Btw I think this change is still in the master branch (416c27aa2b7).

Issue #104, opened 2025-10-03
Cannot enable native tab-bar when `mac-transparent-titlebar` is non-nil


pkryger:
> [!NOTE]
> This option has been disabled for now while a few kinks are worked out.
---

I noticed this when investigating if it would be feasible to change Emacs
application appearance to make text legible when macOS system setting is light,
and frame's background colour is dark (see #99).  In short,
when `mac-transparent-titlebar` is non-nil in a given frame then,
calling `mac-toggle-tab-bar` doesn't do anything, for example compare the frame
displayed after:

* `emacs -Q -eval "(mac-toggle-tab-bar)"`
* `emacs -Q -eval "(progn (set-frame-parameter nil 'mac-transparent-titlebar t) (mac-toggle-tab-bar))"`

My initial investigation revealed that other tab-bar related properties don't
work when the frame has transparent title bar.  E.g. (lines with * are unexpected),

``` emacs-lisp
;; emacs -Q
;; on the initial frame
(set-frame-parameter nil 'mac-transparent-titlebar t)
(mac-frame-tab-group-property nil :selected-frame) ;; => nil *
(mac-frame-tab-group-property nil :frames) ;; => nil *
(mac-set-frame-tab-group-property nil :tab-bar-visible-p t) ;; => no tab-bar is visible *
(mac-set-frame-tab-group-property nil :overview-visible-p t) ;; => no overview is visible *
;; pinch out on trackpad => no overview *

(add-to-list 'default-frame-alist '(mac-transparent-titlebar . t))

;; start a new frame in a new tab
(set-frame-parameter nil 'mac-transparent-titlebar nil)
(mac-set-frame-tab-group-property nil :tab-bar-visible-p t) ;; tab-bar is visible
(mac-handle-new-window-for-tab nil) ;; => a new system tab is shown (with transparent titlebar)
(mac-frame-tab-group-property nil :selected-frame) ;; => nil *
(mac-frame-tab-group-property nil :frames) ;; => nil *
(mac-set-frame-tab-group-property nil :tab-bar-visible-p t) ;; => no tab-bar is visible *
(mac-set-frame-tab-group-property nil :overview-visible-p t) ;; => no overview is visible e*
;; pinch out on trackpad => no overview *


;; start a new frame in a new system window
(make-frame) ;; new system window with a new frame is shown (with transparent titlebar)
(mac-frame-tab-group-property nil :selected-frame) ;; => nil *
[...]

PR #109, opened 2025-10-13
Fix interactions between transparent title bar and native tab bar

pkryger:
This fixes interactions between transparent title bar (a frame with a
non-nil parameter `mac-transparent-tiltebar`) and macOS native tab bar
(setting mac frame tab group properties `:tab-bar-visible-p`
and `:overview-visible-p`).  It is achieved by temporarily disabling
title bar transparency when interacting with `NSWindow`'s functions
related to tab bar.

There's also a few other fixes and improvements:
* Ensure the frame parameter `borders-respect-alpha-background` is
  respected.  The parameter has been introduced in Emacs 31 and there
  was no handler for it in Emacs mac.  I copied the solution, from
  whatever NS port does (just set the parameter value in the frame
  model).
* Avoid setting transparent title bar for tooltip frames.
* Use `YES` and `NO` values for functions that take `BOOL` arguments.
* Update documentation to explicitly note that when transparent title
  bar is used, the frame background should match macOS appearance (light
  or dark).
* Added tests focusing on interactions between transparent title bar and
  native tab bar.

I have tested this change on macOS Sequoia 15.6.1 (M2) and macOS Big Sur
11.7.10 (Intel).

Fixes: #104

### DIFF
--- a/doc/emacs/macport.texi
+++ b/doc/emacs/macport.texi
@@ -540,11 +540,36 @@ title/tool/tab bars look slightly colored with the value of the
 color) if the frame is focused and not in fullscreen.
 
 @vindex mac-transparent-titlebar
- The title bar transparency can be controlled by the value of the
+  The title bar transparency can be controlled by the value of the
 @code{mac-transparent-titlebar} frame parameter.  If it is @code{nil},
 which is the default then title bar uses system color for its
 background.  If it is non-@code{nil} then the frame background color is
 used instead.
+
+It is recommended that the theme applied on the frame with transparent
+title bar matches current macOS appearance.  That is, when macOS
+appearance is Light the theme should have a light colored background.
+Likewise, when macOS appearance is Dark the theme should have a dark
+colored background.
+
+The theme can be automatically changed when macOS appearance changes by
+registering a @code{mac-effective-appearance-change-hook}, for example:
+
+@example
+(add-hook
+ 'mac-effective-appearance-change-hook
+ (lambda ()
+   "Switch theme according to current system setting."
+   (when-let* ((appearance (plist-get (mac-application-state)
+                                      :appearance))
+               (desired-theme (let ((case-fold-search t))
+                                (if (string-match-p "dark" appearance)
+                                    (cadr modus-themes-to-toggle)
+                                  (car modus-themes-to-toggle))))
+               ((not (eq desired-theme
+                         (car custom-enabled-themes)))))
+     (message "Loading theme: %s" desired-theme))))
+@end example
 
 @node Mac Fullscreen
 @section Fullscreen support on the Mac Port
--- a/test/src/mac-transparent-titlebar-tab-bar-tests.el
+++ b/test/src/mac-transparent-titlebar-tab-bar-tests.el
@@ -1,0 +1,163 @@
+;;; mac-transparent-titlebar-tab-bar-tests.el --- Tests for interactions between tranparent title bar and native tab-bar -*- lexical-binding:t -*-
+
+;;; Commentary:
+;;
+;; This tests are designed to run with a visible frame (a GUI window),
+;; as they verify native macOS tab-bar functionality while
+;; `mac-transparent-titlebar' is enabled.  As such there's not
+;; implemeted as ert style tests per se, i.e. not a series of
+;; `deftest's, but rather this file should be evaled as a script, e.g.,
+;;
+;;   emacs -Q -load mac-transparent-titlebar-tab-bar-tests.el
+
+;;; Code:
+
+(require 'ert)
+
+(defun wait-for (secs)
+  "Wait for SECS and force redisplaying the window."
+  (let ((until (+ (float-time) secs))
+        (interval (/ secs 20.0)))
+    (while (< (float-time) until)
+      (sleep-for interval)
+      (force-window-update)
+      (redisplay))))
+
+;; Need some time for the frame to be fully drawn in the screen.
+(wait-for 5)
+
+(setq debug-on-error t)
+
+(defvar frame-stack nil)
+
+;; (0) Enable transparency on current frame and all new frames
+(set-frame-parameter nil 'mac-transparent-titlebar t)
+(add-to-list 'default-frame-alist '(mac-transparent-titlebar . t))
+
+;; (1) Test the initial frame
+(push (should
+       (mac-frame-tab-group-property nil :selected-frame))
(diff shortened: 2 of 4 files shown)


doc/emacs/macport.texi
```diff
@@ -540,12 +540,18 @@ title/tool/tab bars look slightly colored with the value of the
...
 which is the default then title bar uses system color for its
 background.  If it is non-@code{nil} then the frame background color is
 used instead.
 
+It is recommended that the theme applied on the frame with transparent
+title bar matches current macOS appearance.  That is, when macOS
+appearance is Light the theme should have a light colored background.
+Likewise, when macOS appearance is Dark the theme should have a dark
+colored background.
+
```
jdtsmith, line 554: Should we point to `mac-effective-appearance-change-hook` here?

test/src/mac-transparent-titlebar-tab-bar-tests.el
```diff
@@ -0,0 +1,163 @@
...
+(should-not (mac-frame-tab-group-property nil :tab-bar-visible-p))
+(should-not (mac-frame-tab-group-property nil :overview-visible-p))
+
+;; toggle tab-bar on and off
+(mac-set-frame-tab-group-property nil :tab-bar-visible-p t)
+(should (mac-frame-tab-group-property nil :tab-bar-visible-p))
+(mac-set-frame-tab-group-property nil :tab-bar-visible-p nil)
+(should-not (mac-frame-tab-group-property nil :tab-bar-visible-p))
+
+;; TODO: need to find a way to exit overview programatically
```
jdtsmith, line 53: Also to switch focused tabs in the overview by keyboard!

jdtsmith:
Thanks.  I worry a bit about the workarounds being fragile with various versions/improvements; have you tried it on recent MacOS versions?  @simmsb, have a look!

test/src/mac-transparent-titlebar-tab-bar-tests.el
pkryger: That's a good idea. But I am inclined to give it a pass for now - there's a test that switches selected tab by the tab's frame.

doc/emacs/macport.texi
pkryger: Good idea!

pkryger:
> Thanks. I worry a bit about the workarounds being fragile with various versions/improvements; have you tried it on recent MacOS versions?

I don't have access to macOS Tahoe just yet (perhaps in next few weeks or so?). Tried on the two laptops I have access to, one with pretty ancient macOS, the other with "just" one macOS release behind latest and greatest. Perhaps someone could try that?

jdtsmith:
@simmsb could you take a look?

jdtsmith:
What's the status on this, all?

pkryger:
I have been using this patch since I submitted the PR. Although I don't use native tab bar I do use transparent titlebar. I have bot seen any issues. I don't plan any further work to be added unless anyone thinks something can be improved. I guess we're awaiting for @simmsb to review.

jdtsmith:
Thanks for your patience.

shipmints:
In the same vein, it would be good to mirror the upstream emacs frame parameter ```'ns-appearance``` which, when ```'dark``` always shows the frame title bar with a dark background. This is how I use emacs and I never switch system themes.

jdtsmith:
So you have other apps light but Emacs dark?  Craziness.  I think it makes more sense to have the system appearance affect and updated your Emacs theme vs. the other way around.  

@pkryger you looked into this, right?  Is it easy to add, or disruptive?

shipmints:
I've never noticed and it's never concerned me. I mean, between Emacs, Chrome, Terminal/WezTerm, Messages, Preview that's all I use nearly 100% of the time with occasional Notes, Microsoft Office apps. I never use Finder (command line works better for me I ain't gonna click on icons, like ever). I could experiment with a dark system theme just for funsies but I never do on iPhone or iPad either.

pkryger:
> So you have other apps light but Emacs dark?  Craziness.

🤣🤣🤣

> I think it makes more sense to have the system appearance affect and updated your Emacs theme vs. the other way around.  
> 
> 
> 
> @pkryger you looked into this, right?  Is it easy to add, or disruptive?  


I did. It seems that just adding extra frame parameter should be relatively easy - just ensure the appearance is applied in a GUI thread. However, have not followed up on how it would interact with automatic appearance changes. Not sure if NS has a feature like that. Perhaps having a tri-state parameter: `light`, `dark`, and `auto`. Then ignore automatic change when user explicitly set the value?

jdtsmith:
I like it.  With `auto` default.

shipmints:
Makes sense.

P.S. I just tried macOS 26 dark and the stupid mode makes normal Chrome windows the same color as incognito making visual differentiation impossible. I'll skip it.

jdtsmith:
Sounds like a Chrome problem to me.

shipmints:
Of course, but I don't control (and don't want to patch) each app that might have its issues so I'll just deal with my precious Emacs dark appearance on its own with the system's light theme.  If I built Chromium from scratch, perhaps I could patch it.  Every bug report I've made to Chrome/Chromium has been ignored or subsumed in another issue or addressed (unintentionally?) without response.  Same for macOS bugs.  Apple doesn't seem to care either.  macOS (and iOS) have been carrying many of the same bugs since 2001, and introduced truckloads of others along the way.  QA be damned, and users aren't even beta testers since Apple ignores "beta" feedback, so we're just considered sheep.

simmsb:
Heya, sorry for taking ages to get around to this, I've not had a free weekend in a while.

I've had a play around with this on Macos 26, and indeed the tab bar is able to appear when the titlebar is transparent. 


https://github.com/user-attachments/assets/44dc9b10-01ff-406f-95fc-566a412e3a64

simmsb (approved):

shipmints:
Using the term "tab bar" confused me because I thought it meant ```tab-bar``` perhaps we should call it macOS native tab bar to avoid confusing other naive people like me.

pkryger:
> Heya, sorry for taking ages to get around to this, I've not had a free weekend in a while.
>
> I've had a play around with this on Macos 26, and indeed the tab bar is able to appear when the titlebar is transparent. 
> 
> [...]

NW. Thank you for the review a d tests on Tahoe. Do you think you can merge it (I don't have enough privileges to do so)?

PR #86, opened 2025-08-24
Empty GCD drawing queue prior to sleep

jdtsmith:
If GCD is enabled, drawing blocks in the queue during a sleep/wake cycle
may have invalid contexts.  Empty the queue prior to sleep to avoid
this.

* src/macappkit.m ([EmacsController applicationDidFinishLaunching:]):
Add `willSleep' observer.

Hopefully fixes #78.

PR #113, opened 2025-10-25
Fix GCD deadlock on sync queue draw

jdtsmith:
**TLDR:** A queue attempting to flush itself from the "inside" with a call to `dispatch_sync` deadlocks, so don't do that.

Another attempt to fix the dreaded (but rare) GCD draw deadlock, that often (but not always) occurs on waking from sleep.  See #78.  

Here we assume the problem arises from a serial queue deadlock which occurs when a block from the queue running on the main GUI thread attempts to use `dispatch_sync` on the same queue.

So an asynchronous block on the `global_focus_drawing_queue` tries to "flush" the queue, using `dispatch_sync` to quickly wait for it to complete all its submitted blocks.  But if this happens on the main thread, `dispatch_sync` is waiting for the queue to drain, and the currently executing block on the selfsame queue is waiting for `dispatch_sync` to complete: deadlock!  

Apparently this isn't a modern approach to drain GCD queues, but I kept it intact, and simply do nothing to sync if we are calling in to the same queue.

See also #86 for another (failed) attempt.

* src/macappkit.m (set_global_focus_view_frame): set identifier key on
drawing queue for easy identification.
(mac_draw_queue_sync): Do not dispatch empty block if already running
from drawing queue.

### DIFF
--- a/src/macappkit.m
+++ b/src/macappkit.m
@@ -7630,6 +7630,7 @@ static CGContextRef saved_focus_view_context;
 static bool saved_focus_view_modified_p;
 #if DRAWING_USE_GCD
 static dispatch_queue_t global_focus_drawing_queue;
+static const void * kDrawingQueueKey = &kDrawingQueueKey;
 #endif
 
 static void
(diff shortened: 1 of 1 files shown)


jdtsmith:
I have had no recurrences of the issue in #78, so I am again cautiously optimistic that this is the right fix.

PR #115, opened 2025-11-01
Fix setting of window appearance

simmsb:
This PR makes the window appearance correctly change based on the background colour of the frame: when a theme with a light background is loaded, the window appearance changes to 'light', and likewise with themes with dark backgrounds.

When the titlebar is not transparent, the effect is as with changing the system between light and dark mode, so perhaps this isn't an ideal 'fix' as people might have been relying on the title bar appearance always reflecting their system choice?

When the titlebar is transparent, this makes text legible when a light theme is used and the system is in dark mode:

Before: 
Screenshot 2025-11-01 at 14 58 40

After:
Screenshot 2025-11-01 at 15 01 37

I'm not sure why this is needed, perhaps macos changed the source of truth for this parameter, or it was always broken.

### DIFF
--- a/src/macappkit.m
+++ b/src/macappkit.m
@@ -2623,6 +2623,8 @@ static void mac_move_frame_window_structure_1 (struct frame *, int, int);
 	  }
       window.appearanceCustomization.appearance =
 	oldWindow.appearanceCustomization.appearance;
+      window.appearance =
+	oldWindow.appearance;
 
       [oldWindow setDelegate:nil];
       [self hideHourglass:nil];
@@ -5260,6 +5262,8 @@ mac_set_frame_window_background (struct frame *f, unsigned long color)
 	 ? NSAppearanceNameVibrantLight : NSAppearanceNameVibrantDark);
 
       window.appearanceCustomization.appearance =
+	[NSAppearance appearanceNamed:name];
+      window.appearance =
 	[NSAppearance appearanceNamed:name];
     });
 }


jdtsmith:
Thanks. This seems similar to @shipmints' proposal for `mac-appearance` in #109.  This seems to be the `auto` setting being discussed.  Probably worth merging the concepts.

jdtsmith:
Now that you looked at #109, what do you think of the idea of a `mac-appearance` var?  The code this PR touches would presumably need to respect it.

shipmints:
FYI, while helping Prot debug modus-themes 5.0, I reviewed my modus-themes code and I do indeed alter ns-appearance in my custom modus-themes-after-load-theme-hook:

    (when my:on-mac-ns
      (dolist (frame (frame-list))
        (when (display-graphic-p frame)
          (set-frame-parameter frame 'ns-appearance (frame-parameter frame 'background-mode)))))

So I do wind up with the title bar matching the background BUT of my active Emacs theme, not of my system.

jdtsmith:
If there are 3 settings for a putative `mac-appearance` — `light`, `dark`, and `auto` — what is the implicit setting now?

shipmints:
For the sake of consistency, I suggest following the standard values set here, where nil is automatic, not 'auto:

```elisp
(defcustom frame-background-mode nil
...
  :type '(choice (const dark)
                 (const light)
                 (const :tag "automatic" nil)))
```

pkryger:
> If there are 3 settings for a putative `mac-appearance` — `light`, `dark`, and `auto` — what is the implicit setting now?

I guess it's going to be `auto` but we can use it as a standing and code with `nil' meaning automatic setting.

@shipmints Have you tried with this PR applied? Guessing from screenshots and [this snippet](https://github.com/jdtsmith/emacs-mac/blob/f29ff69522b0c1bee400ff9d0c5533e6eb7d5594/src/macappkit.m#L5313-L5316) it should do the right thing (as long as your background is on the right side of the threshold).

jdtsmith:
I meant in the absence of `mac-apperance`, what is it _effectively_?  Always `dark` (like in @simmsb' first screenshot)?

jdtsmith:
I think `nil=auto` is fine.  Does someone want to roll up a PR that includes the addition of a `mac-appearance` (along the lines of `ns-appearance`), implement and document its 3 states, and correct handling of `auto=nil` (which is what this PR I believe does)?

In elisp, if we want to, we can hookup something to `mac-effective-appearance-change-hook` and change the theme too. on system appearance change.  

Is there a "default light/default dark" theme var anywhere?  So many layers to this onion...

Also, @simmsb, one thing I don't get: why doesn't the [NSAppearanceCustomization](https://developer.apple.com/documentation/appkit/nsappearancecustomization?language=objc) do what you wanted already?

jdtsmith:
Anybody want to pull these pieces together and make a `mac-appearance`?  I guess the way it would work:

- `mac-appearance=nil`: automatic, based on the background color.
- `mac-appearance=light`: use light appearance.
- `mac-appearance=dark`: use dark appearance. 

Is the `nil=auto` setting adaptive for existing windows if the theme changes?  

What happens if someone (or we) set up something on `mac-effective-appearance-change-hook` that changes the theme to match the appearance?  Would that create a loop?

jdtsmith:
Hearing none, I applied this for now and we can sort out `mac-appearance` later.

pkryger:
It's somewhere on my TODO list. But I haven't found time to take a look
at it. Thank you for merging this one - I was planning my work on top of
this anyway.

FWIW: I agree with the proposed semantics of `mac-appearance`. I guess
the `mac-effective-appearance-change-hook` should be run when setting
the `mac-appearance` changes effective appearance, correct?

jdtsmith:
> . I guess
> the mac-effective-appearance-change-hook should be run when setting
> the mac-appearance changes effective appearance, correct?

Seems like this could result in an infloop with setting `auto=nil`?  Simplest path is:

1. System appearance changes. 
1. A user switches themes in a `mac-effective-appearance-change-hook`. 
1. The window decoration is auto updated (because `auto`). 

But other paths could lead to loops.  Does a system-wide appearance change do _nothing_ without some opt-in app support?

pkryger:
> > . I guess
> > the mac-effective-appearance-change-hook should be run when setting
> > the mac-appearance changes effective appearance, correct?
> 
> Seems like this could result in an infloop with setting `auto=nil`?  Simplest path is:
> 
> 1. System appearance changes. 
> 1. A user switches themes in a `mac-effective-appearance-change-hook`. 
> 1. The window decoration is auto updated (because `auto`). 
>
> But other paths could lead to loops.  

Good point. I think I was coming from an assumption that user may want to call a hook code function after changing this frame parameter from within a user defined function. Let's postpone running the hook for when someone asks for it. But even then, I would recommend to run the hook function directly from the user defined function.  

> Does a system-wide appearance change do _nothing_ without some opt-in app support?

I think changing Emacs's frame appearance and firing the hook are automatic.

PR #120, opened 2025-12-05
Add mac-appearance frame parameter

pkryger:
This PR adds a new frame parameter `mac-appearance`.  It handles three
values:
* `nil` which means use automatic frame appearance based on current
  system appearance.
* `dark` which forces frame appearance to Vibrant Dark.
* `light` which forces frame appearance to Vibrant Light.

The concept of this parameter bases on `ns-appearance` frame
parameter. It uses Vibrant Light instead of Acqua for `light` setting,
as this is what Emacs-mac has always been doing.

Setting the value to non-`nil` forces the frame (window in macOS parlance) to use selected appearance (Vibrant Light or Vibrant Dark). It may be useful to use it when automatic appearance is not doing the right thing.

Setting the value to nil makes the frame to use automatic frame
appearance, based on frame's background color.

See also #115.

### DIFF
--- a/doc/emacs/macport.texi
+++ b/doc/emacs/macport.texi
@@ -529,31 +529,51 @@ defaults write org.gnu.Emacs Emacs.fringe.attributeStipple alpha:.5
 @code{scroll-bar-background} frame parameter.  If it is @code{nil},
 which is the default, then the frame background color is used instead.
 Its color also affects the appearance of other GUI parts: scroll bars,
-the title bar, the tool bar, (the tab bar on macOS 10.12 and later),
-and popup menus on OS X 10.10 -- macOS 10.13 become light (or dark) if
-the color is considered light (or dark, respectively).  On macOS 10.14
-and later, the appearance of GUI parts other than scroll bars respect
-the global setting, which is available via
+the title bar, the tool bar, (the tab bar on macOS 10.12 and later), and
+popup menus on OS X 10.10 -- macOS 10.13 become light (or dark) if the
+color is considered light (or dark, respectively).  Unless
+@code{mac-appearance} frame parameter is non-@code{nil}, then on macOS
+10.14 and later, the appearance of GUI parts other than scroll bars
+respect the global setting, which is available via the function
 @code{(mac-application-state)}.  On macOS 10.12 and later, the
 title/tool/tab bars look slightly colored with the value of the
 @code{scroll-bar-background} frame parameter (or the frame background
 color) if the frame is focused and not in fullscreen.
 
-@vindex mac-transparent-titlebar
-  The title bar transparency can be controlled by the value of the
-@code{mac-transparent-titlebar} frame parameter.  If it is @code{nil},
-which is the default then title bar uses system color for its
-background.  If it is non-@code{nil} then the frame background color is
-used instead.
+@vindex mac-appearance@r{, a frame parameter}
+  If set to @code{dark} draw this frame's window-system window using the
+``vibrant dark'' theme.  If set to @code{light} draw this frame's
+window-system window using ``vibrant light'' theme.  Otherwise, when set
+to @code{nil} (which is the default) use an automatic appearance based
+on the frame's background color.  When the automatic appearance doesn't
+render a legible frame, the ``vibrant dark'' theme/``vibrant light''
+theme can be used to set the toolbar and scrollbars to a macOS
+Dark/Light appearance when using an Emacs theme with a dark/light
+background.  For example to make Emacs frame appearance to follow system
+appearance the following snipped can be used:
 
-It is recommended that the theme applied on the frame with transparent
-title bar matches current macOS appearance.  That is, when macOS
-appearance is Light the theme should have a light colored background.
-Likewise, when macOS appearance is Dark the theme should have a dark
-colored background.
+@example
+(add-hook
+ 'mac-effective-appearance-change-hook
+ (lambda ()
+   "Change frames appearances according to current system setting."
+   (when-let* ((system-appearance (plist-get (mac-application-state)
+                                             :appearance))
+               (appearance (if (let ((case-fold-search t))
+                                 (string-match "dark" system-appearance))
+                               'dark
+                             'light)))
+     (dolist (frame (frame-list))
+       (unless (eq (frame-parameter frame 'mac-appearance)
+                   appearance)
+         (set-frame-parameter frame 'mac-appearance appearance))))))
+@end example
 
-The theme can be automatically changed when macOS appearance changes by
-registering a @code{mac-effective-appearance-change-hook}, for example:
+  In cases when theme change occurs as a result of system appearance
+change, the automatic frame appearance - based on the theme's background
+may be sufficient.  The theme can be automatically changed when macOS
+appearance changes by registering a
+@code{mac-effective-appearance-change-hook}, for example:
 
 @example
 (add-hook
@@ -563,13 +583,30 @@ registering a @code{mac-effective-appearance-change-hook}, for example:
    (when-let* ((appearance (plist-get (mac-application-state)
                                       :appearance))
                (desired-theme (let ((case-fold-search t))
-                                (if (string-match-p "dark" appearance)
+                                (if (string-match "dark" appearance)
                                     (cadr modus-themes-to-toggle)
                                   (car modus-themes-to-toggle))))
                ((not (eq desired-theme
                          (car custom-enabled-themes)))))
-     (message "Loading theme: %s" desired-theme))))
+     (message "Loading theme: %s" desired-theme)
+     (load-theme desired-theme))))
 @end example
+
+@vindex mac-transparent-titlebar@r{, a frame parameter}
+  The title bar transparency can be controlled by the value of the
+@code{mac-transparent-titlebar} frame parameter.  If it is @code{nil},
+which is the default then title bar uses system color for its
+background.  If it is non-@code{nil} then the frame background color is
+used instead.
+
+  It is recommended that theme applied on a frame with transparent title
+bar matches current macOS appearance (as reported by
+@code{mac-application-state}) when frame parameter
+@code{mac-appearance} is @code{nil} or value of frame parameter
+@code{mac-appearance} when it is non-@code{nil}.  That is, when the
+appearance is Light the theme should have a light colored background.
+Likewise, when the appearance is Dark the theme should have a dark
+colored background.
 
 @node Mac Fullscreen
 @section Fullscreen support on the Mac Port
--- a/src/macappkit.m
+++ b/src/macappkit.m
@@ -4368,6 +4368,32 @@ mac_set_frame_window_modified (struct frame *f, bool modified)
   mac_within_gui (^{[window setDocumentEdited:modified];});
 }
 
+static NSAppearanceName
+automatic_appearance_from_color(unsigned long color)
+{
+  /* This formula comes from frame-set-background-mode in frame.el.  */
+  return (RED_FROM_ULONG (color) + GREEN_FROM_ULONG (color)
+	  + BLUE_FROM_ULONG (color)) >= (int) (0xff * 3 * .6)
+    ? NSAppearanceNameVibrantLight : NSAppearanceNameVibrantDark;
+}
+
+void
+mac_set_frame_window_appearance (struct frame *f, enum mac_appearance_type appearance)
+{
+  EmacsWindow *window = FRAME_MAC_WINDOW_OBJECT (f);
+
+  mac_within_gui (^{
+      NSAppearanceName name =
+	appearance == mac_appearance_system_default
+	? automatic_appearance_from_color (FRAME_BACKGROUND_PIXEL (f))
+	: (appearance == mac_appearance_vibrant_light
+	   ? NSAppearanceNameVibrantLight : NSAppearanceNameVibrantDark);
+      window.appearanceCustomization.appearance =
+	[NSAppearance appearanceNamed:name];
+      window.appearance =
+	[NSAppearance appearanceNamed:name];});
+}
+
 void
 mac_set_frame_window_transparent_titlebar (struct frame *f, bool transparent)
 {
@@ -5315,18 +5341,13 @@ mac_set_frame_window_background (struct frame *f, unsigned long color)
 
   mac_within_gui (^{
       [window setBackgroundColor:[NSColor colorWithEmacsColorPixel:color]];
-      /* This formula comes from frame-set-background-mode in
-	 frame.el.  */
-      NSAppearanceName name =
-	((RED_FROM_ULONG (color) + GREEN_FROM_ULONG (color)
-	  + BLUE_FROM_ULONG (color)) >= (int) (0xff * 3 * .6)
-	 ? NSAppearanceNameVibrantLight : NSAppearanceNameVibrantDark);
-
-      window.appearanceCustomization.appearance =
-	[NSAppearance appearanceNamed:name];
-      window.appearance =
-	[NSAppearance appearanceNamed:name];
-    });
+      if (FRAME_MAC_APPEARANCE (f) == mac_appearance_system_default)
+	{
+	  NSAppearanceName name = automatic_appearance_from_color (color);
+	  window.appearanceCustomization.appearance =
+	    [NSAppearance appearanceNamed:name];
+	  window.appearance =
+	    [NSAppearance appearanceNamed:name];}});
 }
 
 /* Store the screen positions of frame F into XPTR and YPTR.
--- a/src/macfns.c
+++ b/src/macfns.c
@@ -1215,6 +1215,36 @@ mac_set_override_redirect (struct frame *f, Lisp_Object new_value, Lisp_Object o
     }
 }
 
+static enum mac_appearance_type
+mac_set_appearance_1 (struct frame *f, Lisp_Object new_value)
+{
+  enum mac_appearance_type appearance = mac_appearance_system_default;
+  if (EQ (new_value, Qdark))
+    appearance = mac_appearance_vibrant_dark;
+  else if (EQ (new_value, Qlight))
+    appearance = mac_appearance_vibrant_light;
+
+  FRAME_MAC_APPEARANCE (f) = appearance;
+  return appearance;
+}
+
+static void
+mac_set_appearance (struct frame *f, Lisp_Object new_value, Lisp_Object old_value)
+{
+  if ((mac_operating_system_version.major < 10
+       || (mac_operating_system_version.major == 10
+	   && mac_operating_system_version.minor < 10)))
+    return;
+
+  if (!EQ (new_value, old_value))
+    {
+      enum mac_appearance_type appearance = mac_set_appearance_1 (f, new_value);
+      block_input ();
+      mac_set_frame_window_appearance (f, appearance);
+      unblock_input ();
+    }
+}
+
 static void
 mac_set_transparent_titlebar (struct frame *f, Lisp_Object new_value, Lisp_Object old_value)
 {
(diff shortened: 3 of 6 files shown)


doc/emacs/macport.texi
```diff
@@ -529,17 +529,28 @@ defaults write org.gnu.Emacs Emacs.fringe.attributeStipple alpha:.5
...
-title/tool/tab bars look slightly colored with the value of the
-@code{scroll-bar-background} frame parameter (or the frame background
-color) if the frame is focused and not in fullscreen.
-
-@vindex mac-transparent-titlebar
+the title bar, the tool bar, (the tab bar on macOS 10.12 and later), and
+popup menus on OS X 10.10 -- macOS 10.13 become light (or dark) if the
+color is considered light (or dark, respectively).  On macOS 10.14 and
+later, the appearance of GUI parts other than scroll bars respect the
+global setting, which is available via @code{(mac-application-state)},
```
jdtsmith: the function @code{(mac-application-state)}...

doc/emacs/macport.texi
```diff
@@ -529,17 +529,28 @@ defaults write org.gnu.Emacs Emacs.fringe.attributeStipple alpha:.5
...
+unless @code{mac-appearance} frame parameter is non-@code{nil}.  On
+macOS 10.12 and later, the title/tool/tab bars look slightly colored
+with the value of the @code{scroll-bar-background} frame parameter (or
+the frame background color) if the frame is focused and not in
+fullscreen.
+
+@vindex mac-appearance@r{, a frame parameter}
+If set to @code{dark} draw this frame's window-system window using the
```
jdtsmith: and toolbar/etc.?

jdtsmith:
Thanks!  Can you say a bit more about how this relates to @simmsb's #115? I can't quite get the "chain of authority" when it comes to appearance, and how/whether "window decorations" and "toolbar color" differ in treatment.

doc/emacs/macport.texi
```diff
@@ -529,17 +529,28 @@ defaults write org.gnu.Emacs Emacs.fringe.attributeStipple alpha:.5
...
+``vibrant dark'' theme.  If set to @code{light} draw this frame's
+window-system window using ``vibrant light'' theme.  Otherwise, when set
+to @code{nil} (which is the default) use an automatic appearance based
+on the frame's background color.  When the automatic appearance doesn't
+render a legible frame, the ``vibrant dark'' theme/``vibrant light''
+theme can be used to set the toolbar and scrollbars to a dark/light
+appearance when using an Emacs theme with a dark/light background.
```
jdtsmith: Having a hard time parsing this bit.  Is it just explaining in what scenarios the user might want to override the automatic setting?  I think you are just restating the `dark` and `light` settings here?  Maybe not needed.

src/macterm.h
```diff
@@ -339,6 +339,8 @@ struct mac_output
 #define FRAME_MAC_DOUBLE_BUFFERED_P(f) \
   ((f)->output_data.mac->double_buffered_p)
 
+#define FRAME_MAC_APPEARANCE(f) \
+  ((f)->mac_transparent_titlebar)
```
jdtsmith: Wait, shouldn't that be `((f)->mac_appearance)`?

src/macappkit.m
```diff
@@ -4363,6 +4363,32 @@ - (void)windowDidExitTabOverview
...
+
+void
+mac_set_frame_window_appearance (struct frame *f, enum mac_appearance_type appearance)
+{
+  EmacsWindow *window = FRAME_MAC_WINDOW_OBJECT (f);
+
+  mac_within_gui (^{
+      NSAppearanceName name =
+	appearance == mac_appearance_system_default
+	? automatic_appearance_from_color (FRAME_BACKGROUND_PIXEL (f))
```
jdtsmith, line 4388: This seems to make "automatic" defer to the background color in the current frame, _not_ the system setting itself.  Note that `effectiveAppearance` changes happen in ELISP, in `mac-handle-application-effective-appearance-change`.  So I guess the idea here is:

1. System-wide appearance changes occur (by user action in system settings, or automatically e.g. day/night).
2. `mac-handle-application-effective-appearance-change` gets notified (via KVO), and runs the hook `mac-effective-appearance-change-hook`.
3. _If_ the user has a hook in place, it _may_ change the theme on one or more frames.
4. ??? How do existing frames get their appearance updated?

src/macfns.c
```diff
@@ -2547,6 +2577,12 @@ DEFUN ("x-create-frame", Fx_create_frame, Sx_create_frame,
 					  "scrollBarBackground",
 					  "ScrollBarBackground", false);
 
+  tem = gui_display_get_arg (dpyinfo, parms, Qmac_appearance,
+			     NULL, NULL, RES_TYPE_SYMBOL);
+  mac_set_appearance_1 (f, tem);
```
jdtsmith, line 2582: This seems to be the only place where appearance actually gets set, i.e. during `make-frame`?  Is that so?  How are existing frames handled?

jdtsmith (commented):
Thanks for all your work on this.  I haven't had a chance to test it, but just want to make sure I understand the design ideas.

doc/emacs/macport.texi
pkryger: fixed

src/macterm.h
pkryger: It should indeed. Good catch. Fixed.

src/macfns.c
pkryger: AFAIU, there are more places. An existing frame can have its appearance changed via the frame's parameters: `background-color`, `scroll-bar-background`, and `mac-appearance`. The former two are an existing functionality, that uses the color to setup frame appearance. The latter is the new addition (this PR), that can block the other two, if it is not nil for the frame.

So, when a new frame is created it either uses its parent's frame configuration, or `initial-frame-alist`/`default-frame-alist` values. For the existing frame it's up to the user to change the appearance, see [the other comment](https://github.com/jdtsmith/emacs-mac/pull/120/files#r2602187308).

src/macappkit.m
pkryger: Point 1 to 3 is what I understand as well. The `load-theme` sets all existing frames' backgrounds, for example (FWIW: it's `custom-theme-recalc-face` that iterates over all frames):
```
(lldb) breakpoint set -name mac_set_frame_window_background
Breakpoint 3: where = Emacs`mac_set_frame_window_background + 24 at macappkit.m:5335:25, address = 0x0000000100376744
(lldb) run -Q
[... 'c' on a few breakpoints hits...]
[... in an Emacs frame eval: (load-theme 'modus-operandi)...]
* thread #2, name = 'org.gnu.Emacs.lisp-main', stop reason = breakpoint 3.1
    frame #0: 0x0000000100376744 Emacs`mac_set_frame_window_background(f=0x000000084f9b6000, color=16777215) at macappkit.m:5335:25
[...]
(lldb) xbacktrace
(unsigned char *) data = 0x0000000107755160 "internal-set-lisp-face-attribute"
(unsigned char *) data = 0x000000010775526e "set-face-attribute"
(unsigned char *) data = 0x000000010785b118 "apply"
(unsigned char *) data = 0x00000001077552ab "face-spec-set-2"
(unsigned char *) data = 0x00000001077556b8 "face-spec-recalc"
(unsigned char *) data = 0x00000001077556f0 "custom-theme-recalc-face"
(unsigned char *) data = 0x00000001078035a8 "enable-theme"
(unsigned char *) data = 0x0000000107803849 "load-theme"
[...]
(lldb) frame select 5
frame #5: 0x000000010001d564 Emacs`Fmodify_frame_parameters(frame=(struct frame *) $183 = 0x000000084f9b6000, alist=(struct Lisp_Cons *) $180 = 0x0000000170791f00) at frame.c:4086:5
[...]
(lldb) xdebug_print alist
((background-color . "#ffffff"))
```

To confirm that, when I change macOS appearance in System Settings, the frame remains as it was. For example, I started with appearance Light, generated the backtrace above then changed to Dark appearance. The result is as in the screenshot:
Screenshot 2025-12-09 at 11 37 44
The breakpoint has not been hit as a result of changing system appearance.

doc/emacs/macport.texi
pkryger: I have tried to model this paragraph after [the `ns-appearance` one](https://github.com/jdtsmith/emacs-mac/blob/a9f6592fe275774e6c14e3058a4f3488f9914e5e/doc/lispref/frames.texi#L2310-L2317) (which had not the easiest time to do so). I will welcome all suggestions.

doc/emacs/macport.texi
pkryger: Indeed. I thought the ominous `frame's window-system window` covered that aspect. But can change. Will welcome a suggestion though.

pkryger (commented):
I will defer answering to the "chain of responsibility" in lieu of other comments. Please let me know if that's not sufficient, I will try to explain in more detail.

doc/emacs/macport.texi
jdtsmith: Maybe 

> MacOS window decorations

?

jdtsmith:
I think we are on the same page.  So perhaps adding to the documentation how to "close the circuit" using an appearance change hook would help the confused reader.

src/macfns.c
jdtsmith: So where does the "window decoration" e.g. border and toolbar of existing frames get changed?  Is this working for you on existing frames (if you `load-theme` in an appearance change hook)?

doc/emacs/macport.texi
jdtsmith: Maybe just 

> While the default setting of nil should usually give good results, it may be useful to override the automatic appearance determination for certain frame backgrounds/themes.  This can be done for an entire session, or updated in a `mac-effective-appearance-change-hook'. 

But then is the variable `mac-appearance` is updated, this sounds like all frames decorations' are (are they?).

src/macfns.c
pkryger: > So where does the "window decoration" e.g. border and toolbar of existing frames get changed? Is this working for you on existing frames (if you `load-theme` in an appearance change hook)?

It works exactly like that. Gist that should allow to find what exactly going on in [this comment](https://github.com/jdtsmith/emacs-mac/pull/120#discussion_r2602187308). TL;DR: `load-theme` changes all frames' backgrounds, which uses the magic color formula to determine appearance.

doc/emacs/macport.texi
pkryger: To be strinct `mac-appearance` is a frame parameter.  But yeah, in my mind it is kinda like "frame local variable". It's only updated in user code, if the user has a need to override automatic appearance. The automatic appearance has been an always on feature of Emacs-mac for some time now. This PR simply adds an extra control that allows user to chosen appearance for a given frame.

jdtsmith:
Btw I have some travel so will need some time to try this out.  Thanks.

pkryger:
> Btw I have some travel so will need some time to try this out.  Thanks. 

Thanks for heads up and no worries. I think this feature is more for a completeness rather than a critical functionality.

Enjoy your travels!

doc/emacs/macport.texi
```diff
@@ -529,31 +529,51 @@ defaults write org.gnu.Emacs Emacs.fringe.attributeStipple alpha:.5
...
+                                             :appearance))
+               (appearance (if (let ((case-fold-search t))
+                                 (string-match "dark" system-appearance))
+                               'dark
+                             'light)))
+     (dolist (frame (frame-list))
+       (unless (eq (frame-parameter frame 'mac-appearance)
+                   appearance)
+         (set-frame-parameter frame 'mac-appearance appearance))))))
+@end example
```
jdtsmith, line 570: Is another way to do this simply to `load-theme` a dark/light theme, and assume that will DTRT?  Should we provide an example of that?

pkryger: Not sure what do you mean by that...

AFAIU, loading a theme will trigger automatic appearance change, based on frame's background colour. The example provided is for a someone who, don't want to change theme when system appearance changes yet they want their Emacs to follow system appearance. For example a theme with blue based background and a low contrast default face foreground, that looks good when frame decorations  either are light or  are dark.

jdtsmith:
I was planning to test this by back porting to v30 to test out there, but couldn't cherry pick the commits from this PR cleanly.  Does it build on other commits that are only in the experimental master branch?

doc/emacs/macport.texi
jdtsmith: I need to get it built and test it out to formulate my query more clearly :).

pkryger:
> I was planning to test this by back porting to v30 to test out there, but
> couldn't cherry pick the commits from this PR cleanly.  Does it build on other
> commits that are only in the experimental master branch?

Yes... and no... Yes in a sense that it doesn't apply cleanly [1].  And no,
as there should be no dependencies on any other functionality.  Perhaps
with the exception of what has been done in #115, as you probably would
like to test with exactly the same side effects on macOS/Objctive-C
objects.

[1] Conflicts would arise because:
* Documentation in `macport.texi` for `mac-transparent-titlebar` has
  been rewrittent in this PR.
* Frame parameter handlers mapping in array `mac_frame_parm_handlers`
  defined in `macfns.c` is a bit more tricky, as there has been a new
  parameter added in Emacs 31 (see
  798f5ce2dc260baa597d1f43fa6bac84732ab2be).
* Variables, functions and `ifdef`s defined
  for `mac-transparent-titlebar` functionality are right beside the code
  for this PR. Since the `mac-transparent-titlebar` functionality has
  been pulled from Emacs 30 I suspect patching mechanism will struggle
  with applying relevant hunks.

Please let me know if you still want to test within Emacs 30 I can cut a
patch.

jdtsmith:
Thanks.  I should probably get a master build going at some point anyway, so no worries.

jdtsmith:
Thanks again for working on this.  I was able to build and test this PR on our master branch.  It works well.  There is, however, some odd interaction with `mac-transparent-titlebar`.  

If you start with `Dark` system-wide appearance (which I guess only matters to get a dark theme set), and then eval:

```elisp
(set-frame-parameter nil 'mac-transparent-titlebar t) ; evaluate for the new frame
(set-frame-parameter nil 'mac-appearance 'light)

(list (frame-parameter nil 'mac-transparent-titlebar)
      (frame-parameter nil 'mac-appearance)) ; t light

(mac-next-tab-or-toggle-tab-bar nil) ; then hit the `+' tab
(set-frame-parameter nil 'mac-transparent-titlebar t) ; evaluate for the new frame
(set-frame-parameter nil 'mac-appearance 'light)
```

so that both tabs have frame parameters `transparent=t` and `appearance=light`, you get:

image

Maybe this is a question for @simmsb: why doesn't "transparent" apply to other unselected tabs?  Maybe there's no simple way to handle this, other than to recommend avoiding explicit `mac-appearance` when using transparent titlebar.  Thoughts?  

Then if I load a theme like `modus-vivendi`, and ensure `t light` params as above for both frames (tabs), I get something odd.  One tab looks normal, and one tab is "fully dark":

image

I.e. the behavior of transparent titlebar is inconsistent.  In one case it takes the frame background (pure black in vivendi), in the other tab it takes something else.  `scroll-bar-background` perhaps (which differs for some reason between the tabs/frames)?

jdtsmith:
@pkryger (+@simmsb) could you take a look when you have a moment?  It's possible this native tabs + transparent titlebar + explicit appearance combination is as good as we can get it, but I'm not expert enough in each to know.

pkryger:
> @pkryger (+@simmsb) could you take a look when you have a moment? It's possible this native tabs + transparent titlebar + explicit appearance combination is as good as we can get it, but I'm not expert enough in each to know.

I hope I will have some time this week to take a look at this.

jdtsmith:
@pkryger any luck?

pkryger:
I have not been looking at this for a while. I may have some spare cycles in a couple of weeks.

pkryger:
It took me longer than anticipated to get back to this. I rebased the code on the most recent master, fixed one goof (use `EmacsWindow` instead of `NSWindow` in `mac_set_frame_window_appearance`) and rebuilt it. 

I can no longer reproduce the following issue on macOS 26.4.1 (IIRC I was able to render similar effect on earlier macOS 26, but I no longer have access to it). I can see the tab bar blinking when switching between tabs that have different value of `mac-transparent-titlebar`, but the white background title is not there any more.

> If you start with `Dark` system-wide appearance (which I guess only matters to get a dark theme set), and then eval:
> 
> ```emacs-lisp
> (set-frame-parameter nil 'mac-transparent-titlebar t) ; evaluate for the new frame
> (set-frame-parameter nil 'mac-appearance 'light)
> 
> (list (frame-parameter nil 'mac-transparent-titlebar)
>       (frame-parameter nil 'mac-appearance)) ; t light
> 
> (mac-next-tab-or-toggle-tab-bar nil) ; then hit the `+' tab
> (set-frame-parameter nil 'mac-transparent-titlebar t) ; evaluate for the new frame
> (set-frame-parameter nil 'mac-appearance 'light)
> ```
> 
> so that both tabs have frame parameters `transparent=t` and `appearance=light`, you get:
> 
> image
> Maybe this is a question for @simmsb: why doesn't "transparent" apply to other unselected tabs? Maybe there's no simple way to handle this, other than to recommend avoiding explicit `mac-appearance` when using transparent titlebar. Thoughts?


I am speculating here, but the following my steam from the fact that `srcoll-bar-background`/`scroll-bar-foreground` frame parameters are not changed on all frames when loading theme. For example in  `Emacs -Q`:

``` elisp
;; make a new tab
(let ((mac-frame-tabbing t))
  (make-frame))
;; go back to the first tab
(load-theme 'modus-operandi)
(mapcar (lambda (p) (frame-parameter nil p))
	'(scroll-bar-background scroll-bar-foreground))
;; ("#f2f2f2" "#9f9f9f")

;; go to the new tab
(mapcar (lambda (p) (frame-parameter nil p))
	'(scroll-bar-background scroll-bar-foreground))
;; (nil nil)
	
```

Note: when theme is loaded from the new tab, the first tab has `srcoll-bar-background`/`scroll-bar-foreground` set. I think this is a separate issue though.

> Then if I load a theme like `modus-vivendi`, and ensure `t light` params as above for both frames (tabs), I get something odd. One tab looks normal, and one tab is "fully dark":
> 
> image
> I.e. the behavior of transparent titlebar is inconsistent. In one case it takes the frame background (pure black in vivendi), in the other tab it takes something else. `scroll-bar-background` perhaps (which differs for some reason between the tabs/frames)?

jdtsmith:
Sorry for the long delay in response, thanks for looking into it.  Have you been using this setting in your build with good results?

I was reminded of this PR by a [discussion](https://www.reddit.com/r/emacs/comments/1uepn3x/emacs_with_ghosttylike_liquid_glass_effect/) of a "glass" appearance setting for NS.  I proposed perhaps hooking into our concept of "system" (= tunable glass) using `mac-appearance`.  Not proposing that for now, but do you think the current approach could accommodate some glass appearance options in the future?

pkryger:
> Sorry for the long delay in response, thanks for looking into it. 

NW. Sorry about delays on my side too.

> Have you been using this setting in your build with good results?

I have been using a build with the feature compiled in, but personally I don't use native tabs (I use `tab-bar` instead), so I have not been using other settings. I have not noticed any problems that I could relate to it. I've added this feature mostly for parity with NS and (I think) someone has asked for it.

> I was reminded of this PR by a [discussion](https://www.reddit.com/r/emacs/comments/1uepn3x/emacs_with_ghosttylike_liquid_glass_effect/) of a "glass" appearance setting for NS. I proposed perhaps hooking into our concept of "system" (= tunable glass) using `mac-appearance`. Not proposing that for now, but do you think the current approach could accommodate some glass appearance options in the future?

I have not checked how the feature was implemented in the NS, but I guess something like that can be added to mac build.

jdtsmith:
Are there any issues that prevent merging?  Take a look at #146 to see if that's a concern.  Otherwise, if you don't mind retargeting the new `emacs-31` branch we can get this in.  Apologies for the long delay.

PR #124, opened 2026-01-09
Quote $0 in Emacs.sh to handle paths with spaces

MeetThePatel:
I ran into this while setting up a second Emacs on my system called `Emacs Canary.app`, which surfaced issues with unquoted paths containing spaces. This change makes the startup script handle that case correctly.


jdtsmith:
Thanks.

PR #135, opened 2026-06-11
mac accessibility: avoid keymap access from GUI thread in unsafe windows

liaowang11:
## Problem

On macOS, an external accessibility client inspecting the menu bar can hang Emacs hard (requiring force-quit). I captured two `sample` runs 78s apart plus a system hang report — all show the **same** wedged stack, so it's a true deadlock, not slow progress.

The GUI/main thread faults (SIGSEGV) here:

```
mshMIGPerform → _XCopyAttributeValue → _AXXMIGCopyAttributeValue
  → -[NSMenu(Accessibility) _openForInspection:] → _simulateOpening:
    → _enableItems → _enableItem: → targetForAction:to:from:
      → -[EmacsController respondsToSelector:]
        → is_action_selector → access_keymap → access_keymap_1   ← SIGSEGV
          → handle_sigsegv → deliver_fatal_thread_signal → __sigsuspend   ← GUI thread parked forever
```

Meanwhile the Lisp thread receives the forwarded fatal signal and gets stuck in the orderly-shutdown path (`deliver_fatal_signal → terminate_due_to_signal → shut_down_emacs → kill_buffer_processes`), unable to complete because the GUI thread is suspended mid-AppKit-AX-callback (holding CF/AppKit/AX locks). Net result: a permanent hang instead of a crash.

## Root cause

`is_action_selector` runs on the GUI thread (synchronous accessibility callbacks reach it via `-respondsToSelector:` / menu validation) and walks the Lisp keymaps in `mac-apple-event-map` (`get_keymap`, `access_keymap`, `intern`). When the Lisp thread is mid-mutation — during GC, or the `inhibit-quit`/select-emulation window — that access can fault.

This is the same class of hazard addressed in 627940726de ("mac accessibility: try harder to prevent buffer access issues"), which guarded the buffer-*content* accessors with `(poll_suppress_count == 0 && !NILP (Vinhibit_quit)) || gc_in_progress`. The menu/keymap path was left unguarded.

## Fix

Bail out of `is_action_selector` (returning `NO`) during those same unsafe windows, mirroring the existing buffer-access guards. Outside those windows (real menu-action dispatch) behavior is unchanged; during them, the accessibility/menu probe sees the selector as temporarily not-an-action instead of faulting.

### DIFF
--- a/src/macappkit.m
+++ b/src/macappkit.m
@@ -13241,6 +13241,15 @@ is_action_selector (SEL selector)
 {
   NSString *name = NSStringFromSelector (selector);
 
+  /* This runs on the GUI thread, including from synchronous
+     accessibility callbacks (e.g. menu inspection via
+     -respondsToSelector:).  Walking the keymaps below touches Lisp
+     data that the Lisp thread may be altering, so bail out during the
+     windows when such access is unsafe.  */
+  if ((poll_suppress_count == 0 && !NILP (Vinhibit_quit))
+      || gc_in_progress)
+    return NO;
+
   /* The selector name is of the form `ACTIONNAME:' ?  */
   if (NSMaxRange ([name rangeOfString:@":"]) == [name length])
     {


jdtsmith:
Thanks.  I've also seen various accessibility lockups.  I think modern macOS versions make a lot more use of those requests.  I've concluded that the current design in emacs-mac is not really possible to make fully safe.  I'm working on a new version that has many differences, but Accessibility handling is one of them.  From the notes:

> Mac accessibility features are very robust, but they put a heavy load on apps.  The accessibility server is synchronous, and expects immediate answers to requests for information about text geometry and contents.  On most Mac apps, the GUI thread is the exclusive owner of this kind of information, but in Emacs, the LISP thread is.  emacs-mac used to use heuristics to try to safely "sneak a peek" from the GUI thread at the underlying gap-buffer data, but this didn't always work and could cause crashes.  Now, when LISP is idle and processing tasks, GUI coordinates with it to gather the relevant buffer information (with a timeout), eliminating thread sync accessibility issues.  This does mean that sometimes when LISP is busy, certain accessibility information will not be available, but in practice this is rare.

Have you tested this fix to see that it works reliably?

liaowang11:
> Thanks. I've also seen various accessibility lockups. I think modern macOS versions make a lot more use of those requests. I've concluded that the current design in emacs-mac is not really possible to make fully safe. I'm working on a new version that has many differences, but Accessibility handling is one of them. From the notes:
> 
> > Mac accessibility features are very robust, but they put a heavy load on apps.  The accessibility server is synchronous, and expects immediate answers to requests for information about text geometry and contents.  On most Mac apps, the GUI thread is the exclusive owner of this kind of information, but in Emacs, the LISP thread is.  emacs-mac used to use heuristics to try to safely "sneak a peek" from the GUI thread at the underlying gap-buffer data, but this didn't always work and could cause crashes.  Now, when LISP is idle and processing tasks, GUI coordinates with it to gather the relevant buffer information (with a timeout), eliminating thread sync accessibility issues.  This does mean that sometimes when LISP is busy, certain accessibility information will not be available, but in practice this is rare.
> 
> Have you tested this fix to see that it works reliably?

I have been testing it actively, there seems to be some performance degradation with it, but I'm not quite sure if it's related to this patch. A new version that works more reliably would be much appreciated, it's ok to close this as I think this is still just a workaround.

jdtsmith:
I can't imagine this would slow anything down.  Returning `nil` is common practice when you can't answer accessibility requests now.  I'll leave this open as a reminder, and hope that when my new version drops you'll give it some solid Accessibility testing.

liaowang11:
The performance seems mostly related to new native compilations I think, it's working smoothly lately. Can't wait to try the new version and I will definitely give solid testing for that.

PR #143, opened 2026-08-30
mac: support GUI and TTY frames from daemons

srnnkls:
## Summary

- restore terminal clients for Mac GUI instances
- select the Mac backend when a daemon or terminal frame is current
- register the `Mac` display before window-system initialization
- start AppKit event processing after a daemon initializes its first Mac terminal

## Failure

A frame-less daemon loaded `mac-win`, but `window-system` was nil and the
`Mac` display mapping was installed only during GUI initialization.  Its
first `emacsclient -c` therefore failed with `Don't know how to interpret
display "Mac"` and then fell back to a terminal client.  The daemon also
kept `inhibit_window_system` set after initializing the Mac terminal, so
AppKit events would not be pumped.

The preceding multi-TTY commit restores the Emacs 29 behavior needed for
GUI and terminal clients to coexist in one Mac instance.

## Validation

- `make -j"$(sysctl -n hw.logicalcpu)"`
- `make -C test lisp/server-tests` — 7/7 passed
- `make -C test lib-src/emacsclient-tests` — 2/2 passed
- isolated foreground daemon with a private home and socket directory
- terminal client before GUI initialization and after GUI initialization
- first GUI frame creation, resize, redraw, focus, close, and reopen
- repeated evaluation pings and clean daemon shutdown

### DIFF
--- a/lisp/server.el
+++ b/lisp/server.el
@@ -955,7 +955,7 @@ This handles splitting the command if it would be bigger than
     ;; can not make X frames.
     (cond ((featurep 'ns-win)
 	   (setq w 'ns display "ns"))
-	  ((eq window-system 'mac)
+	  ((featurep 'mac-win)
 	   (setq w 'mac display "Mac"))
           (t
       ;; FIXME! Not sure what this was for, and not sure how it should work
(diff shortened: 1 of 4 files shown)


jdtsmith:
Can you retarget and test on the new `emacs-mac-31` branch?

jdtsmith:
Also, is this equivalent to the patches proposed [here](https://ylluminarious.github.io/2019/05/23/how-to-fix-the-emacs-mac-port-for-multi-tty-access/)?

srnnkls:
Retargeted.

> Also, is this equivalent to the patches proposed [here](https://ylluminarious.github.io/2019/05/23/how-to-fix-the-emacs-mac-port-for-multi-tty-access/)?

Yes, added an attribution.

PR #144, opened 2026-08-30
mac: prevent frame churn from starving dispatch work

srnnkls:
Stacked on #143.  The final commit is the dispatch-starvation fix; this
PR will shrink to that commit when #143 lands.

## Summary

- disable AppKit close animations before Emacs frame teardown
- run front-to-back backing copies on a private serial queue
- run the asynchronous `pselect` worker on a separate private serial queue

## Failure

Rapid frame creation and deletion started one blocking AppKit animation
worker per closed frame.  At the 64-thread soft limit, no global worker
remained for Emacs work needed by the GUI thread.  Samples captured the
GUI waiting for a backing copy or select completion while the Lisp thread
waited for the GUI.

Moving only the backing copy exposed the same starvation in the select
worker and image loading.  Disabling close animations removes the worker
leak; the private queues keep the two blocking synchronization paths off
the shared global pool.

## Validation

- `make -j"$(sysctl -n hw.logicalcpu)"`
- `make -C test lisp/server-tests` — 7/7 passed
- `make -C test lib-src/emacsclient-tests` — 2/2 passed
- five regular-GUI trials totaling 250 visible frame create/redraw/delete cycles
- 100 equivalent cycles against a frame-less daemon after its first GUI frame
- final sample reported no dispatch soft-limit saturation
- repeated evaluation pings, GUI close/reopen, and clean shutdown

### DIFF
--- a/src/macappkit.m
+++ b/src/macappkit.m
@@ -2724,6 +2724,8 @@ static void mac_move_frame_window_structure_1 (struct frame *, int, int);
      causes emacsView to receive drawRect: before closing a tabbed
      window on macOS 10.12.  It is too late to remove the view in the
      windowWillClose: delegate method, so we remove it here.  */
+  /* AppKit close animations block one global dispatch worker per frame.  */
+  emacsWindow.animationBehavior = NSWindowAnimationBehaviorNone;
   [emacsView removeFromSuperview];
   [emacsWindow close];
 }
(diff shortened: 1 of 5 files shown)


jdtsmith:
Thanks for the contribution.  Most of this code has been supplanted in a new version I'm working on.  

Can you clearly describe the issue you are having?  You can't create 64 frames at a time?  Over what time frame?

srnnkls:
I was benchmarking emacsclient frame creation latency, one frame at a time. 

Two runs with a sample size of 30, one with gc enabled, one with gc disabled. Accumulating close animation workers exposed the bug accidentally.

src/macappkit.m
```diff
@@ -16650,6 +16664,9 @@ - (void)sound:(NSSound *)sound didFinishPlaying:(BOOL)finishedPlaying
    select emulation.  */
 static dispatch_source_t mac_select_dispatch_source;
 
+/* This must stay runnable when AppKit saturates global dispatch workers.  */
+static dispatch_queue_t mac_select_queue;
+
 /* Command to execute in the GUI thread after the run loop is
```
jdtsmith, line 16670: Why do you need a static queue here?

srnnkls: The global queue is the problem: it’s width-limited, so under AppKit frame churn `pselect` can get stuck waiting for a worker while the GUI waits on `pselect`. A private static queue avoids that starvation and is static because there’s only one process-wide select emulation. The backing-copy queue is now per-`EmacsBacking`, so copies across frames can still run in parallel.

PR #150, opened 2026-09-13
Introduce nix flake for development builds

what-the-functor:
This flake will make it easier for those using Nix to build development builds and releases, quicker than the Nixpkgs release cycle.


jdtsmith:
Sorry, but I don't think it's prudent for us to get into packaging details like these, which are better handled downstream, similar to the various packagers for the NS build.

